PR #90, opened 2014-06-02
Add false sharing protection to RingBuffer and SingleProducerSequencer 

nitsanw:
Added pre/post padding to RingBuffer.
Added pre/post padding elements to array, correct sequence calculation accordingly, use Unsafe for array access to eliminate resulting  pad addition overhead (and a minor win to eliminating boundary check)
Inline cache fields into Sequencer and pre/post pad them.

### DIFF
--- a/src/main/java/com/lmax/disruptor/SingleProducerSequencer.java
+++ b/src/main/java/com/lmax/disruptor/SingleProducerSequencer.java
@@ -19,22 +19,34 @@ import java.util.concurrent.locks.LockSupport;
 
 import com.lmax.disruptor.util.Util;
 
+abstract class SingleProducerSequencerPad extends AbstractSequencer
+{
+    protected long p1, p2, p3, p4, p5, p6, p7;
+    public SingleProducerSequencerPad(int bufferSize, WaitStrategy waitStrategy)
+    {
+        super(bufferSize, waitStrategy);
+    }
+}
+abstract class SingleProducerSequencerFields extends SingleProducerSequencerPad
+{
+    public SingleProducerSequencerFields(int bufferSize, WaitStrategy waitStrategy)
+    {
+        super(bufferSize, waitStrategy);
+    }
 
+    /** Set to -1 as sequence starting point */
+    protected long nextValue = Sequence.INITIAL_VALUE;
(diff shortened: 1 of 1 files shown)


src/main/java/com/lmax/disruptor/RingBuffer.java
```diff
@@ -24,14 +27,39 @@
  *
  * @param <E> implementation storing the data for sharing during exchange or parallel coordination of an event.
  */
-public final class RingBuffer<E> implements Cursored, DataProvider<E>
+abstract class RingBufferPad
 {
-    public static final long INITIAL_CURSOR_VALUE = Sequence.INITIAL_VALUE;
+    protected long p1, p2, p3, p4, p5, p6, p7;
```
xuxiaoxia: I cann't understand well,why did you add pre/post padding to RingBuffer? Please do some explain.Thanks!

mikeb01: To prevent false sharing: http://mechanical-sympathy.blogspot.co.nz/2011/07/false-sharing.html

xuxiaoxia: Hi  mikeb01,We know that
class Sequence{
	protected long p1, p2, p3, p4, p5, p6, p7;
	protected volatile long value;
	protected long p9, p10, p11, p12, p13, p14, p15;
}
the p1~p15 in Sequence class above is used to protect "value" field from false sharing,I can understand this well.
But I can understand which field(or fields) the padding in the RingBuffer.java protect from false sharing. Thanks!

mikeb01: The RingBuffer fields:

    private final long indexMask;
    private final Object[] entries;
    protected final int bufferSize;
    protected final Sequencer sequencer;

Could end up sharing a cache line with some other data.

xuxiaoxia: OK,Thank you very much!

zhupengcheng:
I think in RingBuffer the Buffer_pad (line 51) isn't  need 128 bytes, 64 bytes is ok. is this Right?thx

mikeb01:
See comment on issue #158

PR #108, opened 2015-02-24
More generics around usages of ExceptionHandler.java

hsrinara:
We've pushed the generics for exception handler around a bit more so that the disruptor<T> and exceptionHandler<T> tie up. 

Things to look out for in your review:
 - we've converted some usages of <?> to <T> 
 - we've convinced ourselves that the cast in ExceptionHandlerSetting  with <T> will work out ok - but this bit felt a little dodgy.


mikeb01:
Cheers, merged.

PR #115, opened 2015-05-14
Alternative usage of event poller

brimzi:
Was trying to use the EventPoller using the example currently given but I realized that it is much slower since the benefits of batching are lost. So I wrote a wrapper around it and the results were very impressive. Please let me know of what you think.


mikeb01:
Nice example, merged.

PR #125, opened 2015-09-07
Add Disruptor.setDefaultExceptionHandler API

### LINKED ISSUE #122: ExceptionHandler isn't propagated by Disruptor to existed WorkerPool after set
ExceptionHandlers set to Disruptor after initializing WorkerPool will not be propagated to WorkerPool

```
disruptor = new Disruptor<>(Event::new, RING_SIZE, executor, ProducerType.SINGLE, new BlockingWaitStrategy());

disruptor.handleExceptionsWith(new FatalExceptionHandler());
disruptor.handleEventsWith(new Handler<Event>()).handleEventsWithWorkerPool(new AnotherHandler<Event>(), new AnotherHandler<Event>());
```

Will fail with NullPointer exception in case of exception in AnotherHandler (exception handler in WorkerPool is null), while on this code

```
disruptor = new Disruptor<>(Event::new, RING_SIZE, executor, ProducerType.SINGLE, new BlockingWaitStrategy());

disruptor.handleEventsWith(new Handler<Event>()).handleEventsWithWorkerPool(new AnotherHandler<Event>(), new AnotherHandler<Event>());
disruptor.handleExceptionsWith(new FatalExceptionHandler());
```

will handle exception correctly.
I think this code samples should lead to the same result and state.

ajsutton:
…on handler for all event processors and worker pools created by Disruptor regardless of whether they were created before or after the call to setDefaultExceptionHandler. This fixes #122

### DIFF
--- a/src/main/java/com/lmax/disruptor/dsl/Disruptor.java
+++ b/src/main/java/com/lmax/disruptor/dsl/Disruptor.java
@@ -250,7 +250,7 @@ public class Disruptor<T>
         {
             throw new IllegalStateException("Mixing calls to handleExceptionsWith and setDefaultExceptionHandler is not supported.");
         }
-        ((ExceptionHandlerWrapper<T>)this.exceptionHandler).setDelegate(exceptionHandler);
+        ((ExceptionHandlerWrapper<T>)this.exceptionHandler).switchTo(exceptionHandler);
     }
 
     /**
--- a/src/main/java/com/lmax/disruptor/dsl/Disruptor.java
+++ b/src/main/java/com/lmax/disruptor/dsl/Disruptor.java
@@ -248,7 +248,7 @@ public class Disruptor<T>
         checkNotStarted();
         if (!(this.exceptionHandler instanceof ExceptionHandlerWrapper))
         {
-            throw new IllegalStateException("Mixing calls to handleExceptionsWith and setDefaultExceptionHandler is not supported.");
+            throw new IllegalStateException("setDefaultExceptionHandler can not be used after handleExceptionsWith");
         }
         ((ExceptionHandlerWrapper<T>)this.exceptionHandler).switchTo(exceptionHandler);
     }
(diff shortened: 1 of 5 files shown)


src/main/java/com/lmax/disruptor/dsl/ExceptionHandlerWrapper.java
```diff
@@ -0,0 +1,32 @@
+package com.lmax.disruptor.dsl;
+
+import com.lmax.disruptor.ExceptionHandler;
+import com.lmax.disruptor.FatalExceptionHandler;
+
+public class ExceptionHandlerWrapper<T> implements ExceptionHandler<T>
+{
+    private ExceptionHandler<? super T> delegate = new FatalExceptionHandler();
+
+    public void setDelegate(final ExceptionHandler<? super T> exceptionHandler)
```
SamBarker: would ```updateHandler``` or ```switchToHandler``` make the intent clearer?

src/main/java/com/lmax/disruptor/dsl/Disruptor.java
```diff
@@ -236,6 +236,24 @@ public void handleExceptionsWith(final ExceptionHandler<? super T> exceptionHand
...
+     *
+     * @param exceptionHandler the exception handler to use.
+     */
+    @SuppressWarnings("unchecked")
+    public void setDefaultExceptionHandler(final ExceptionHandler<? super T> exceptionHandler)
+    {
+        checkNotStarted();
+        if (!(this.exceptionHandler instanceof ExceptionHandlerWrapper))
+        {
+            throw new IllegalStateException("Mixing calls to handleExceptionsWith and setDefaultExceptionHandler is not supported.");
```
Spikhalskiy: We can do disruptor().setDefaultHandler().handleExceptionsWith(), so, maybe, not "mixing", but something like "callToDefaultExceptionHandler is not supported and useless after exceptionHandler has overridden by EventHandler specific ExceptionHandler using handleExceptionWith()"

PR #139, opened 2015-12-17
Short-circuit logic should be used in boolean contexts

kirill-vlasov:
This pull request is focused on resolving occurrences of sonar issue squid:S2178 - Short-circuit logic should be used in boolean contexts

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S2178
Please let me know if you have any questions.

Kirill Vlasov

### DIFF
--- a/src/main/java/com/lmax/disruptor/EventPoller.java
+++ b/src/main/java/com/lmax/disruptor/EventPoller.java
@@ -53,7 +53,7 @@ public class EventPoller<T>
                     nextSequence++;
 
                 }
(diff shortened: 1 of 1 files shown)


mikeb01:
I'm going to close this one.  The code as it currently stands is intentional.  Using the non-short-circuit logic allows for the comparisons to be executed in parallel by the CPU whereas the short-circuit logic introduces instruction dependencies.  When benchmarking some time ago this appeared to be faster.

PR #145, opened 2016-02-09
Fix build Issues

rajatvig:
Source Code compatibility should be set to 1.7 and also some CheckStyle issues need to be fixed

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,8 +63,8 @@ idea.module {
     scopes.TEST.plus += [ configurations.perfCompile ]
 }
 
-sourceCompatibility = 1.6
-targetCompatibility = 1.6
+sourceCompatibility = 1.7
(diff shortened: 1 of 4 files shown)


mikeb01:
Thank you for the PR, but I've fixed it slightly differently.  Source should stay 1.6 and I removed the nonsense checkstyle rule.

rajatvig:
Cool. That works. Quick question, though why should the source be 1.6? It's been dead a while.

mikeb01:
Compatibility, I don't want to change the source version on a minor release.  Internally to the Disruptor there is not much benefit to the language features in 7 & 8.  Users can happily use it with the later JDKs as the APIs have been designed with FunctionInterfaces in mind.  I'm planning on moving to Java 9 for version 4, which will use VarHandles instead of Unsafe.

PR #152, opened 2016-04-04
fix MultiBufferBatchEventProcessor endOfBatch handling bug

zergood:
There is a bug in `MultiBufferBatchEventProcessor`: when it reads a first available message in a ringbuffer it always set `endOfBatch` to false because `previous` would be -1 and `available` would be 0.

I've made a sample to illustrate this https://github.com/zergood/disruptor-sample

### DIFF
--- a/src/perftest/java/com/lmax/disruptor/support/MultiBufferBatchEventProcessor.java
+++ b/src/perftest/java/com/lmax/disruptor/support/MultiBufferBatchEventProcessor.java
@@ -57,8 +57,6 @@ public class MultiBufferBatchEventProcessor<T>
         }
 
         final int barrierLength = barriers.length;
-        final long[] lastConsumed = new long[barrierLength];
-        fill(lastConsumed, -1L);
 
         while (true)
         {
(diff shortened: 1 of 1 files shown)


src/perftest/java/com/lmax/disruptor/support/MultiBufferBatchEventProcessor.java
```diff
@@ -57,8 +57,6 @@ public void run()
         }
 
         final int barrierLength = barriers.length;
-        final long[] lastConsumed = new long[barrierLength];
-        fill(lastConsumed, -1L);
```
mikeb01: What role does the lastConsumed array serve.  It seems to only be declared and used here no where else.  It appears to be redundant.

zergood: `lastConsumed` is used only in 61 line. I think it is redundant.

mikeb01: Sorry, brain explosion I just realised you had deleted them.

Issue #138, opened 2015-12-09
Dead lock observed in BlockingWaitStrategy in Log 4J


SampathK:
We have seen this behavior in during high load. Where Logging Got Stropped and Application Went to not responsive state.
log4J Version : 2.2  Disruptor Version :  3.3.2
Ring Buffer Size : 128
Producer(Multiples Threads) and Consumer Threads(Single Thread As per Log 4J Configuration) Started Waiting on each other.

Here is the  one of the Trace from Thread Dump:

Producer : 
"[ACTIVE] ExecuteThread: '7' for queue: 'weblogic.kernel.Default (self-tuning)'" TIMED_WAITING
	sun.misc.Unsafe.park(Native Method)
	java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:349)
	com.lmax.disruptor.MultiProducerSequencer.next(MultiProducerSequencer.java:136)
	com.lmax.disruptor.MultiProducerSequencer.next(MultiProducerSequencer.java:105)
	com.lmax.disruptor.RingBuffer.publishEvent(RingBuffer.java:444)
	com.lmax.disruptor.dsl.Disruptor.publishEvent(Disruptor.java:256)
	org.apache.logging.log4j.core.async.AsyncLogger.logMessage(AsyncLogger.java:285)
	org.apache.logging.log4j.spi.AbstractLogger.logMessage(AbstractLogger.java:722)
	org.apache.logging.log4j.spi.AbstractLogger.logIfEnabled(AbstractLogger.java:693)
        org.apache.logging.log4j.jcl.Log4jLog.debug(Log4jLog.java:81)

Consumer Thread : 
 
"AsyncLogger-1" waiting for lock      java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject@5d972983 WAITING
sun.misc.Unsafe.park(Native Method)
java.util.concurrent.locks.LockSupport.park(LockSupport.java:186)
java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2043)
com.lmax.disruptor.BlockingWaitStrategy.waitFor(BlockingWaitStrategy.java:45)
com.lmax.disruptor.ProcessingSequenceBarrier.waitFor(ProcessingSequenceBarrier.java:55)
com.lmax.disruptor.BatchEventProcessor.run(BatchEventProcessor.java:123)
[...]

PR #159, opened 2016-06-19
Add LiteTimeoutBlockingWaitStrategy

remkop:
Log4j 2 uses the TimeoutBlockingWaitStrategy as the default for its Async Loggers (see related issues https://github.com/LMAX-Exchange/disruptor/issues/138 and https://issues.apache.org/jira/browse/LOG4J2-1221 ).

Recent discussion on the Mechanical Sympathy mailing list suggested that some performance improvement may be possible under lower workloads where the consumer thread often falls asleep. In this scenario, if multiple producer threads get additional work they can cooperate in waking up the consumer. Before trying to wake up the consumer, they can first check if another producer has already started to wake up the consumer, and avoid taking the lock if this is the case. This is exactly what the LiteBlockingWaitStrategy does.

This pull request is to add a wait strategy that combines LiteBlockingWaitStrategy with TimeoutBlockingWaitStrategy.

Question: the LiteBlockingWaitStrategy javadoc says that "this wait strategy should be considered experimental as I have not full proved the correctness of the lock elision code". Is this still the case? It looks like the projectreactor.io project used to use this wait strategy. Did they ever give you feedback on any issues with this strategy?

### DIFF
--- a/src/main/java/com/lmax/disruptor/LiteTimeoutBlockingWaitStrategy.java
+++ b/src/main/java/com/lmax/disruptor/LiteTimeoutBlockingWaitStrategy.java
@@ -1,0 +1,83 @@
+package com.lmax.disruptor;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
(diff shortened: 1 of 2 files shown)


mikeb01:
I'll have a look at merging this later this week.  The LiteBlockingWaitStrategy should work, but I haven't been able to absolutely convince myself that it is correct.  It was based of an optimisation from the Rust implementation of the Disruptor.

mikeb01:
Merged manually.

PR #161, opened 2016-07-04
Fix bug in factor value check in Histogram

ablanchet:
I've just implemented the value check for the `factor` parameter in `getUpperBoundForFactor` as it was described in the method doc.

### DIFF
--- a/src/main/java/com/lmax/disruptor/collections/Histogram.java
+++ b/src/main/java/com/lmax/disruptor/collections/Histogram.java
@@ -331,9 +331,9 @@ public final class Histogram
      */
     public long getUpperBoundForFactor(final double factor)
     {
-        if (0.0d >= factor || factor >= 1.0d)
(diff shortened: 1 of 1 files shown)


mikeb01:
Hi.  I'll merge the commit, but I seriously recommend against using this implementation of a Histogram.  I'm going to mark it as deprecated.  Please use [HdrHistogram](https://github.com/HdrHistogram/HdrHistogram), it is better every way.

ablanchet:
Hi @mikeb01 thank you for your quick answer. I'm not really using this implementation of Histogram, actually I'm working on refreshing the .NET port of the disruptor and just scanning through all the files and found this minor detail.

If you're marking this class as deprecated it's ok to close this PR and kill this Histogram implementation for good :)

mikeb01:
That would be best.  There is a .NET version of the HdrHistogram available too.

PR #165, opened 2016-07-21
Proposition: Add a pause after CAS failure in MultiProducerSequencer

ocoanet:
While analyzing the performance of the .NET port of the Disruptor, I noticed that the CAS in MultiProducerSequencer was the main contention point for 3P1C scenarios. The aggressive busy-spin on the CAS seems to be harm the overall throughput.

By adding a pause with `LockSupport.parkNanos(1)` when the CAS fails I was able to improve the throughput of 3P1C scenarios. On my machine, the ops/sec for ThreeToOneSequencedThroughputTest increase from about 13M to about 21M. The ThreeToOneSequencedBatchThroughputTest performance was also improved. I did not see any change in the other scenarios.

However this pause might have a negative impact on the fairness of the sequencer, or on the max latencies.

What are your thoughts?

### DIFF
--- a/src/main/java/com/lmax/disruptor/MultiProducerSequencer.java
+++ b/src/main/java/com/lmax/disruptor/MultiProducerSequencer.java
@@ -144,6 +144,10 @@ public final class MultiProducerSequencer extends AbstractSequencer
             {
                 break;
             }
+            else
+            {
+                LockSupport.parkNanos(1);
+            }
         }
         while (true);
 


mikeb01:
I don't think I will merge this change.  It will add to much latency overhead.  On Linux a call to LockSupport.parkNanos will take around 60µs which is quite a lot for a CAS failure.  It is also worth noting that we don't want to over optimise for this benchmark.  In a real scenario there is likely to be more work between operations and the contention is less likely.  Worth noting that Intel actually changed they way their hardware works such that the heavily contented case (like this benchmark) would perform worse in favour of more real world scenarios.  Have a look at:

http://mechanical-sympathy.blogspot.co.nz/2011/09/adventures-with-atomiclong.html
http://mechanical-sympathy.blogspot.co.nz/2013/01/further-adventures-with-cas.html

ocoanet:
Ok, thank you for the detailed comment and for the links!

Just so you know, I experimented with a busy-spin (`Thread.SpinWait` in .NET, I do not know the equivalent for Java) to create shorter pauses and the scenario throughput was also improved.

But I understand that the contention on the CAS is a micro-benchmarking issue. The performance tests could be commented to explain this behavior. It seems to me that an improvement in `MultiProducerSequencer` could have a negative impact on the 3P1C scenarios because it would generate more contention.

mikeb01:
There's an equivalent coming in Java 9.  I'm planning on making the Disruptor Java 9 specific for version 4.

PR #170, opened 2016-08-11
Initial version of ParkingWaitStrategy.

RollingHeavy:
Created a wait strategy that uses LockSupport.park() instead of the ReentrantLock.

### DIFF
--- a/src/main/java/com/lmax/disruptor/ParkingWaitStrategy.java
+++ b/src/main/java/com/lmax/disruptor/ParkingWaitStrategy.java
@@ -1,0 +1,73 @@
+package com.lmax.disruptor;
+
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.AbstractQueuedSynchronizer;
+import java.util.concurrent.locks.LockSupport;
+
+/**
+ * Parking strategy that spins then parks the thread. The goal of this srategy
+ * is to avoid the object creation that occurs when using subclasses of
+ * {@link AbstractQueuedSynchronizer}. This strategy will also only work with a
+ * single processing thread.
+ * 
+ * <p>
+ * This strategy can be used when throughput and low-latency are not as
+ * important as CPU resource.
+ */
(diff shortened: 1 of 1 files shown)


mikeb01:
What happens if there is more than one thread waiting?

RollingHeavy:
Hi Michael,    If you have more than one EventHandler, I think that it won't work (if I'm reading the LMAX code correctly, which is always suspect!).  I believe that since all processors use the same WaitStrategy that, in this case, all calling threads could park(), which would in turn overwrite the thread in the AtomicStamoedReference.  So, only the last parked thread would ever unpark.  We never use multiple EventHandlers so it's not a use case we've run through.  Let me try a few tests with it, and make some modifications so it would work in that case.  My goal is to still park without creating more trash than necessary.
Thanks for the email,-Brian 

    On Thursday, August 11, 2016 4:56 PM, Michael Barker <notifications@github.com> wrote:
 

 What happens if there is more than one thread waiting?—

mikeb01:
If you only ever use one event handler, then it is okay in your use case.  You may find that is optimal for you.  It that situation it would be best to keep the implementation in your own code base - perfectly valid thing to do, we designed the wait strategies so that users could have their own.

I think in order to merge it in, it would need to handle multiple event handlers and be demonstrably faster than the BlockingWaitStrategy.

RollingHeavy:
Thanks Michael, I appreciate it.  Yeah I don't have performance numbers vs blocking, this was for garbage reduction.  I could make it work with multiple threads if you're interested, if not no worries.  I just wanted to give something back, the disruptor is a nice library.
Thanks,-Brian 

    On Thursday, August 11, 2016 7:08 PM, Michael Barker <notifications@github.com> wrote:
 

 If you only ever use one event handler, then it is okay in your use case. You may find that is optimal for you. It that situation it would be best to keep the implementation in your own code base - perfectly valid thing to do, we designed the wait strategies so that users could have their own.I think in order to merge it in, it would need to handle multiple event handlers and be demonstrably faster than the BlockingWaitStrategy.—

PR #178, opened 2016-11-08
Improve code coverage of Histogram collection up to 100%

DenisIstomin:
Made code code coverage via IntelliJ IDEA. Improved code coverage of Histogram collection up to 100%.


mikeb01:
Hi, thank you for the patch (will merge later), but there is not a lot of value in working on the Histogram.  Most people wanting well function histograms are using the https://github.com/HdrHistogram.  You'll see that I've moved over to it in the latency tests.

PR #196, opened 2017-05-16
Add TwoArg and ThreeArg forms of publishEvent to Disruptor

kscaldef:


kscaldef:
Hi, what's the prognosis for getting this merged?  It's not urgent since one can easily just perform such operations on the ring buffer directly, but it seems like a nice convenience to have all the versions of publishEvent proxied.

mikeb01:
Hi, sorry haven't had a chance to look at it yet.  Will have a look later today.

PR #216, opened 2018-01-26
Fixed a broken link to the technical paper.

staslev:
The link to the technical paper in the main link section is broken.
Luckily a valid link appears just below, so the fix is trivial.


staslev:
Duplicate of #207, closing.

PR #225, opened 2018-03-27
Add batch size to performance tests

MendelMonteiro:
We've recently added the batch percentage and average batch size to the disruptor-net performance tests.  We've found that it allows us to quickly see how the effects of batching impact the throughput figures.  I.e. the counter-intuitive effect that a slower consumer can lead to higher throughput as the rate of batching increases.

### DIFF
--- a/src/perftest/java/com/lmax/disruptor/AbstractPerfTestDisruptor.java
+++ b/src/perftest/java/com/lmax/disruptor/AbstractPerfTestDisruptor.java
@@ -16,8 +16,6 @@
 package com.lmax.disruptor;
 
 
-import sun.misc.Perf;
-
 public abstract class AbstractPerfTestDisruptor
 {
     public static final int RUNS = 7;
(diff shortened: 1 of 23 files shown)


mikeb01:
I had a quick look.  They seem to impact the performance of the tests by around 10-20%.  Also there is still some formatting issues that is causing the semaphore build to break.  I need to have a more detailed look at the changes.

MendelMonteiro:
I've changed the implementation to use the BatchStartAware interface which improves the performance as the branching statement is not in the tight loop in the batch event processor.  [This is the performance comparison](https://gist.github.com/MendelMonteiro/e1332fa6659df596b01d3bb425436311) between the branch and master as run on my machine (YMMV).

As for the formatting issues, there are many formatting issues that are not related at all to the changes in this PR.  I don't think I've introduced any new ones.

mikeb01:
Cheers.

MendelMonteiro:
While trying to use the `BatchStartAware` interface on the .NET port I've noticed that the `EventPoller` doesn't detect if the `EventHandler` implements the interface.  Is this something that we'd like to add to the EventPoller.

If so I'll add it in another PR...

mikeb01:
Yes, feel free to submit.

PR #236, opened 2018-10-18
about producer's next(n) method

zhangchaodlut:
next(n) will be parked forever when n is bigger than bufferSize property


mikeb01:
Thank you.

PR #239, opened 2018-10-23
close #233

### LINKED ISSUE #233: BUFFER_PAD << REF_ELEMENT_SHIFT always 128 ?
Source code can be found [here](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/RingBuffer.java#L53), it seems that `BUFFER_PAD << REF_ELEMENT_SHIFT` is always equal to 128, if so why not use literal directly?

jiacai2050:
use 128 instead of (BUFFER_PAD << REF_ELEMENT_SHIFT)


mikeb01:
Hi, I'm going to close this one.  See my comment on #233 .

PR #259, opened 2019-03-13
Don't call BatchStartAware#onBatchStart(0) when SequenceBarrier#waitFor() has no progress

mzheravin:
According the WaitStrategy and Sequencer contracts it is possible for SequenceBarrier#waitFor() to return without making any progress. And it is happening sometimes for MultiProducerSequencer indeed. 
In that case BatchStartAware#onBatchStart is not expected to be called.


mikeb01:
Hi, thank you for the patch.  Please could you include a small unit test to show the new functionality?

mzheravin:
Hi Michael, I've added a test that verifies the new functionality. Thanks.

PR #270, opened 2019-07-30
Introduce new data structure for faster shutdown()

joshbooks:
Repeated addAll operations on Sequence[]s in shutdown() were causing excessive allocations, using a linked list to join the arrays yielded instead of allocating and copying seems to significantly speed things up. Happy to post YourKit results if that would be helpful.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Sun Jul 07 12:07:34 PDT 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.1-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.1-all.zip
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,4 +1,3 @@
-#Sun Jul 07 12:07:34 PDT 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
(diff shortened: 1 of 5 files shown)


grumpyjames:
Thanks for the PR!

I have some thoughts:

1. This is a large change compared to its impact. Can you tell us more about the problems excessive allocation during shutdown are causing you? In general we waive our requirements to be efficient when we're shutting down. We're either bringing the entire application to a halt; not usually a performance sensitive area, or about to create a new disruptor; much more expensive than shutting down an existing one.

**Summary**: We would like to know more about the problems this is causing you before adding code to fix it :-)

2. If we did want to make an improvement here, I would consider a different approach. If I take a look at `ConsumerRepository::getLastSequenceInChain`, it only has a single (in library) usage, and that's to support `Disruptor::hasBacklog`. We could instead just get `Disruptor` to delegate `hasBacklog` to `ConsumerRepository`. In this world, we will perform no allocations _and_ there is less code; woohoo. I'd then consider deprecating/removing `getLastSequenceInChain` (although that has versioning implications).

Thanks again for getting in touch - looking forward to hearing more about your use case :-)

joshbooks:
You are very welcome! 
Sure, I’d be very happy to tell you about the problem I’m trying to solve, I’m profiling an application in which I use the disruptor and the excessive allocations are causing noise in my profiler. I could just introduce a pause before shutdown and manually stop profiling then but that would mean I couldn’t get up and go get coffee while I wait, which is clearly unacceptable.

Your suggested solution sounds awesome and I’d prefer to solve it that way if you’re open to it, I just wasn’t sure what the intention was for this interface going forward.

Thanks for the review @grumpyjames

grumpyjames:
Ok, cool - would definitely accept a PR based on those requirements; developer coffee retrieval ergonomics are very important to us ;-). 

Looking forward to seeing PR v2!

P.S If you're occupied elsewhere, let us know and we'll find some time to put this change in ourselves.

joshbooks:
That is most excellent! I will find some time to get that in this weekend.

PR #274, opened 2019-08-26
Delegate hasBacklog to ConsumerRepository

joshbooks:
Stop doing unnecessary allocations in hasBacklog by delegating to ConsumerRepository rather than having ConsumerRepository pass back a Collection made of concatenated arrays.

### DIFF
--- a/src/main/java/com/lmax/disruptor/dsl/ConsumerRepository.java
+++ b/src/main/java/com/lmax/disruptor/dsl/ConsumerRepository.java
@@ -60,6 +60,30 @@ class ConsumerRepository<T> implements Iterable<ConsumerInfo>
         }
     }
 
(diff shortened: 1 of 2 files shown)


joshbooks:
@grumpyjames here's the pr v2 with requsted changes. My only concern about this version is that the `shouldGetLastEventProcessorsInChain` test is now using a deprecated interface and using the new `hasBacklog` interface would result in making much weaker assertions. Not sure if you're concerned about that, but if you are I'd be happy to write a new test and/or get rid of the old one.

ocoanet:
The `ConsumerRepository` class does not seem to be public, is it really necessary to keep the `getLastSequenceInChain` method? Also, I suppose that it would be better to add new tests? I can make a new PR if that can help.

Issue #234, opened 2018-09-27
Modularize Disruptor JAR


Jeff-G:
Add a module-info.java file to the project, in order to provide official support for Java 9+.  Without this, users are unable to run jlink against dependent Java code.

mikeb01:
I have some outstanding work to support JDK 9 with a multi-version jar.  That is stuck behind some internal work to update out build system.  When that is available I'll look at the module-info.java.  I want to continue to support older JDKs for a period of time.  If you are familiar with Java modules and want to contribute, that would be welcomed.

Jeff-G:
I'm not 100% certain what packages you want to be part of the public API, but I would recommend something similar to the following:

    // module-info.java, added to disruptor/src/main/java/
    module com.lmax.disruptor {
        requires java.logging;
        requires java.management;
        requires jdk.unsupported; // This provides sum.misc.Unsafe

        exports com.lmax.disruptor;
        // Also export com.lmax.disruptor.dsl and com.lmax.disruptor.util,
        // if they are part of the public interface
    }

mikeb01:
How would that work if I wanted to do a build that would still work with JDK 7.

Jeff-G:
I have read from a number of sources that module-info.java *should* be ignored by Java 8.  Nowhere could I find a claim for what happens when a modular JAR is used with Java 7.  I would probably just add it in and test in all versions of Java that you intend to support.

Jeff-G:
I have tested with the module file indicated above in JDK 11, and it appears to work when used with Log4j2.  I have also discovered that there are some tools (Eclipse IDE and the Javadoc utility) that don't fully support multi-release JAR files.  If you convert this to a multi-release JAR file, please also add the following to the manifest file in the JAR:

Automatic-Module-Name: com.lmax.disruptor

mikeb01:
Multi-release jars currently break our build tool, which is using a legacy version of ASM.  I'm working on updating it, but it is queued behind a number of Yak shaving exercises.  I'll look at including this once we start supporting JDK 9+.

skay2k:
Analysing disruptor-3.3.2.jar  with jdeps shows some incompatibilities with JDK 11 ; do you know if a new release is comming ? :

`disruptor-3.4.2.jar -> jdk.unsupported
   com.lmax.disruptor.MultiProducerSequencer          -> sun.misc.Unsafe                                    JDK internal API (jdk.unsupported)
[...]

PR #279, opened 2019-10-05
Add Automatic-Module-Name to MANIFEST.MF

matteodri:
This is the first step in the process of modularizing the Disruptor JAR #234.
Adding Automatic-Module-Name in the manifest file makes the library usable as a Java automatic module without moving the library itself to Java 9 (or later) or creating a module-info.java.
This allows the Disruptor when loaded as an automatic module to get its name from the Automatic-Module-Name entry in MANIFEST.MF instead of being inferred from the filename. This second option is dangerous as filenames are far from being unique and stable.
This post explains the point of having an Automatic-Module-Name defined: http://branchandbound.net/blog/java/2017/12/automatic-module-name/

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,6 +32,7 @@ ext {
     teamName = 'LMAX Disruptor Development Team'
     siteUrl = 'http://lmax-exchange.github.com/disruptor'
     sourceUrl = 'git@github.com:LMAX-Exchange/disruptor.git'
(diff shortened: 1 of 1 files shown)


rgoers:
Log4j 2 uses the disruptor. One of the goals of Log4j 2 3.x is to fully modularize all of our deliverables, so we need a released version that includes the automatic module name. When can we expect this in a release?

PR #300, opened 2020-07-26
Switched numbers for variables (test smell)

leofernandesmo:
It is a test refactoring.
The idea is to remove the smell Magic Number.
This smell occurs when a test method contains unexplained and undocumented numeric literals as parameters or as values to identifiers. These magic values do not sufficiently indicate the meaning/purpose of the number.

### DIFF
--- a/src/test/java/com/lmax/disruptor/MultiProducerSequencerTest.java
+++ b/src/test/java/com/lmax/disruptor/MultiProducerSequencerTest.java
@@ -22,20 +22,28 @@ import org.junit.Test;
 
 public class MultiProducerSequencerTest
 {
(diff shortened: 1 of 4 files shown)


leofernandesmo:
Reason of fail: https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/README.md#replacing-use-of-gradles-deprecated-osgi-plugin

> "In Gradle 5.0, the OSGi Plugin was deprecated. If you used the OSGi Plugin and want to switch to the Bnd Gradle Plugin, you will need to modify your build scripts."

sea36:
Hard-coded numbers aren't automatically bad.

This change makes the tests harder to understand.

PR #304, opened 2020-09-06
enable GitHub Actions CI

sullis:
Continuous Integration build using GitHub Actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,28 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


vlsi:
What do you think of moving wrapper validation to its own workflow file, and trying https://github.com/burrunan/gradle-cache-action instead of `./gradlew` call?

In other words:

```
- uses: burrunan/gradle-cache-action@v1
  name: Build Disruptor
  with:
    job-id: jdk${{ matrix.java }}
    arguments: build
    gradle-version: wrapper
```

sullis:
> What do you think of moving wrapper validation to its own workflow file, and trying https://github.com/burrunan/gradle-cache-action instead of `./gradlew` call?

I prefer to use GitHub Actions that are published by established/trusted authors.

Generally speaking, I try to to use GitHub's core Actions only.

Palmr:
> I prefer to use GitHub Actions that are published by established/trusted authors.

:+1: 

Thank you for the contribution

PR #305, opened 2020-09-16
Using https for maven link

dainv91:


Palmr:
This request probably should have been merged.
But since it was opened the `build.sh`, where the improvement was made, has been removed from the project.

PR #309, opened 2020-10-26
Updating links to the Ring Buffer blog post.

trishagee:
The old mechanitis blog site is super deprecated, and will soon be gone. I've updated the link to point to the new site.


trishagee:
Erm. I broke the build but I'm not sure why, apart from that I don't think this docs branch has Gradle!

Palmr:
Hi @trishagee , thanks for the link update.
Sorry about the CI getting a little too keen here. Will merge this request as I think we're safe.

PR #310, opened 2020-10-31
Put github pages in master branch under docs folder

Palmr:
The current gh-pages branch has failed CI in the past due to not having Gradle or the project sources.
This commit brings the contents of the gh-pages branch in to the docs folder of master which can be set as the source of GitHub Pages.

The theme has changed since the old one is no longer easily available.

This also sets the Javadoc destination directory to be under docs, so re-generating Javadoc needs no further action to be in the correct place for the GitHub Pages link.

If this gets merged we will also need to update settings to have this take effect.

Tested on my branch and all links appear to work: https://palmr.github.io/disruptor/


lyncmi07:
Awesome thank you

lyncmi07 (approved):
It worked on your branch so I'm going to assume you know what you're doing here

PR #311, opened 2020-11-01
Java 11 | Gradle 6 | Junit 5

eliquinox:


Palmr:
This pull request is a bit too large to review.
It would be wise to have separate changes for Java 11, then gradle upgrade, then junit update with vintage and finally one or more changes for migrating junit tests.

PR #312, opened 2020-11-06
Non functional cleanup

Palmr:
Remove seemingly unused bash scripts:

 - `build.sh`
 - `runme.sh`
 - `runoffheap.sh`

And moved changelog from readme to its own file


sea36 (approved):

sea36 (commented):
Happy with the removal of build.sh

runme.sh and runoffheap.sh provide flags for running performance tests, so might be useful to keep

Palmr:
> runme.sh and runoffheap.sh provide flags for running performance tests, so might be useful to keep

I have taken a look and:
 - `runme.sh` seems to just have GC logging flags and a `sed` running after to print an attempt at a pause time histogram
 - `runoffheap.sh` runs some other "Throughput Tests" which aren't JMH, with a flag for setting max Direct Memory and runs the tests with perf logging

These might be handy but given the state of the perf testing in general on this project currently, I don't feel too inclined to keep them around.

In terms of keeping them, git history will keep their memory alive :)

grumpyjames:
I'm happy to merge this and point/laugh at @Palmr when he inevitable forgets the lessons of runme.sh and runmeoffheap.sh.

PR #313, opened 2020-11-06
Changed source and target versions to Java 11

eliquinox:
* Changed source and target versions to Java 11

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [ 8, 11 ]
+        java: [ 11 ]
     name: Java ${{ matrix.java }}
     steps:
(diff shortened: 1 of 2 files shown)


Palmr:
You're going to want to remove Java 8 from the GitHub Actions with this change, CI failing to build using JDK8

README.md
```diff
@@ -19,6 +19,10 @@ LMAX Development Team
 
 ## Changelog
 
+### 3.4.3
```
Palmr, line 22: This change is definitely NOT a minor point release. It would break compatibility with 99% of users of this library.
Revert the changelog change as this will not go out in a release any time soon, not until 4.0

Palmr (changes requested):
Changelog shouldn't have been changed

Palmr (approved):

Palmr:
Seems alright, paving the way towards VarHandle and Junit 5

PR #314, opened 2020-11-07
Add a JMH Benchmark

Palmr:
All current performance tests appear to be written before JMH existed and relies on building a separate jar which gives results in a non-standard format.
We should use JMH and leverage the way it works and the profilers it comes with instead.

This is just testing the waters, converting `SimplePerformanceTest` only, before jumping in and writing more JMH benchmarks.
I am not well versed in writing JMH benchmarks so review welcome.

(Note: we can also pass the JMH run some JVM args to make it similar to `runme.sh`)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,7 @@ ext {
 
 sourceSets {
     perf.java.srcDir file('src/perftest/java')
+    jmh
 }
 
 eclipse.classpath.plusConfigurations += [ sourceSets.perf.compileClasspath ]
@@ -55,6 +56,9 @@ dependencies {
     checkstyle 'com.puppycrawl.tools:checkstyle:8.12'
     testCompile 'junit:junit:4.12'
     perfCompile 'org.hdrhistogram:HdrHistogram:2.1.10'
+    jmhCompile project
+    jmhCompile 'org.openjdk.jmh:jmh-core:1.26'
+    jmhAnnotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess:1.26'
 }
 
 idea.module {
@@ -176,6 +180,32 @@ task perfJar(type: Jar) {
     from sourceSets.perf.output
     from sourceSets.test.output
     with jar
+}
+
+task jmh(type: JavaExec, description: 'Executing JMH benchmarks', group: 'performance') {
+    classpath = sourceSets.jmh.runtimeClasspath
+    main = 'org.openjdk.jmh.Main'
+
+    def profilers = ['pauses']
+    def format = 'json'
+    def resultFile = file("build/reports/jmh/result.${format}")
+    resultFile.parentFile.mkdirs()
+
+    args '-rf', format
+    args '-rff', resultFile
+    args '-foe', 'true' //fail-on-error
+    args '-v', 'NORMAL' //verbosity [SILENT, NORMAL, EXTRA]
+    profilers.each {
+        args '-prof', it
+    }
+    args '-jvmArgsPrepend', '-Xmx256m'
+    args '-jvmArgsPrepend', '-Xms256m'
+}
+
+task jmhProfilers(type: JavaExec, description:'Lists the available profilers for the jmh task', group: 'performance') {
+    classpath = sourceSets.jmh.runtimeClasspath
+    main = 'org.openjdk.jmh.Main'
+    args '-lprof'
 }
 
 wrapper {
--- a/build.gradle
+++ b/build.gradle
@@ -200,6 +200,7 @@ task jmh(type: JavaExec, description: 'Executing JMH benchmarks', group: 'perfor
     }
     args '-jvmArgsPrepend', '-Xmx256m'
     args '-jvmArgsPrepend', '-Xms256m'
+    args '-jvmArgsPrepend', '-XX:MaxDirectMemorySize=1g'
 }
 
 task jmhProfilers(type: JavaExec, description:'Lists the available profilers for the jmh task', group: 'performance') {
--- a/src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
+++ b/src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
@@ -22,6 +22,7 @@ import org.openjdk.jmh.runner.options.Options;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
 
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
 @BenchmarkMode(Mode.AverageTime)
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
@@ -30,20 +31,21 @@ import java.util.concurrent.TimeUnit;
 public class SimpleBenchmark
 {
     private RingBuffer<SimpleEvent> ringBuffer;
-    private SimpleEventHandler eventHandler;
     private Disruptor<SimpleEvent> disruptor;
+    private AtomicLong eventsHandled;
 
     @Setup
     public void setup()
     {
+        eventsHandled = new AtomicLong();
+
         disruptor = new Disruptor<>(SimpleEvent::new,
                 Constants.RINGBUFFER_SIZE,
                 DaemonThreadFactory.INSTANCE,
                 ProducerType.SINGLE,
                 new BusySpinWaitStrategy());
 
-        eventHandler = new SimpleEventHandler();
-        disruptor.handleEventsWith(eventHandler);
+        disruptor.handleEventsWith(new SimpleEventHandler(eventsHandled));
 
         ringBuffer = disruptor.start();
     }
@@ -52,6 +54,8 @@ public class SimpleBenchmark
     @OperationsPerInvocation(Constants.ITERATIONS)
     public void publishSimpleEvents()
     {
+        eventsHandled.set(0);
+
         for (int i = 0; i < Constants.ITERATIONS; i++)
         {
             long sequence = ringBuffer.next();
@@ -60,7 +64,7 @@ public class SimpleBenchmark
             ringBuffer.publish(sequence);
         }
 
-        while (eventHandler.lastSeenSequence % Constants.ITERATIONS != Constants.ITERATIONS - 1)
+        while (eventsHandled.get() != Constants.ITERATIONS)
         {
             Thread.yield();
         }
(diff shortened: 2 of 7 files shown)


src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @TearDown
+    public void tearDown() {
+        disruptor.shutdown();
+    }
+
+    public static void main(String[] args) throws RunnerException {
+        Options opt = new OptionsBuilder()
+                .include(SimpleBenchmark.class.getSimpleName())
+                .threads(4)
```
grumpyjames, line 79: This isn't my favourite number of threads. Why four here?

Palmr: Because that's what it was when I copied & pasted from the JMH samples main method... Unfortunately I don't think I can put your favourite number of threads in there, 0, so might have to settle for 1 instead.

grumpyjames:
I think "convert" is doing a lot of work here - the new benchmark does things in a considerably tidier way than SimplePerformanceTest once did. I like tidier, to be clear.

src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+    @OperationsPerInvocation(Constants.ITERATIONS)
+    public void publishSimpleEvents() {
+        for (int i = 0; i < Constants.ITERATIONS; i++) {
+            long sequence = ringBuffer.next();
+            SimpleEvent simpleEvent = ringBuffer.get(sequence);
+            simpleEvent.setValue(i);
+            ringBuffer.publish(sequence);
+        }
+
+        while (eventHandler.lastSeenSequence % Constants.ITERATIONS != Constants.ITERATIONS - 1)
```
grumpyjames, line 63: I can't work out why we need the modulo nonsense here - iterations < ring buffer size, can just wait for an exact match?

src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
```diff
@@ -0,0 +1,80 @@
...
+        {
+            Thread.yield();
```
grumpyjames, line 65: I suspect we could have an entire conversation just about this line. Does the jmh manual have much to say about this sort of test, where we have to have thread co-ordination because we're testing a concurrent data structure?

src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
Palmr: JMH appears to call `publishSimpleEvents` repeatedly, and I only make the ringbuffer once in setup. So the first iteration it works, second iteration it failed because the `lastSeenSequence` was 2x `Constants.ITERATIONS`

I did have a go at putting an `AtomicLong` in the benchmark which the event handler could increment and the benchmark could check and reset each iteration. Just using the sequence and a modulo seemed like the faster way and kept the benchmark tight?

src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
grumpyjames: There's some interesting stuff from Nitsan, [here](http://psy-lob-saw.blogspot.com/p/jmh-related-posts.html) which I'll work my way through. If all else fails I'll have to go and read the full Shipilev lowdown.

Palmr: > jmh manual

I'm not sure such a thing exists? Appears to be some samples in a GitHub project or seeing opinion pieces from various blogs

build.gradle
```diff
@@ -178,6 +182,32 @@ task perfJar(type: Jar) {
     with jar
 }
 
+task jmh(type: JavaExec, description: 'Executing JMH benchmarks', group: 'performance') {
```
grumpyjames, line 185: *sideshow bob rake noises*. This is probably fine, but I'm the last person to ask, gradle is completely greek to me.

src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
grumpyjames: From the quickest of squizzes I think I am prepared to say that there isn't really a consensus, and I get reasonable numbers out of jmh when I run this locally so we're probably good.

build.gradle
Palmr: It's a task, so you can `./gradlew jmh` to run the benchmarks.
It will execute some Java, specifcally the main mentioned in the body.
A nice description for when you wonder what the tasks do, and a group so that all performance-related tasks are grouped together.

Simple!

src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
Palmr: Out of curiosity, what numbers do you get? As a fresh-faced disruptor-maintainer I wasn't sure what would actually count as "reasonable".

On my i5-6600K I get:

```
Benchmark                                            Mode  Cnt    Score   Error  Units
SimpleBenchmark.publishSimpleEvents                  avgt    5   11.831 ± 0.611  ns/op
SimpleBenchmark.publishSimpleEvents:·pauses          avgt   37  365.617             ms
SimpleBenchmark.publishSimpleEvents:·pauses.avg      avgt         9.882             ms
SimpleBenchmark.publishSimpleEvents:·pauses.count    avgt        37.000              #
SimpleBenchmark.publishSimpleEvents:·pauses.p0.00    avgt         7.692             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.50    avgt         9.912             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.90    avgt        11.564             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.95    avgt        12.306             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.99    avgt        13.500             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.999   avgt        13.500             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.9999  avgt        13.500             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p1.00    avgt        13.500             ms
```

src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
grumpyjames: Oh - interesting - so we don't necessarily get a new RingBuffer/handler per invoc.

With my pedant hat on (ok, fine, my default hat) - isn't that condition also true before we do the publish? Should we require the handler's sequence to have increased in addition?

We also appear to be reading a plain old non-volatile long from SimpleEventHandler which is being written to from another thread. I can see why we don't want to pay the cost of updating a volatile long, and yes, the benchmark does terminate, but still, I am unkeen. Might be worth pairing on an alternative here, unless I am in possession of an inappropriate stick end?

src/jmh/java/com/lmax/disruptor/SimpleBenchmark.java
grumpyjames: Intel(R) Xeon(R) CPU           E5450  @ 3.00GHz

```
Benchmark                                            Mode  Cnt     Score   Error  Units
SimpleBenchmark.publishSimpleEvents                  avgt    5    15.888 ± 2.733  ns/op
SimpleBenchmark.publishSimpleEvents:·pauses          avgt  382  1767.211             ms
SimpleBenchmark.publishSimpleEvents:·pauses.avg      avgt          4.626             ms
SimpleBenchmark.publishSimpleEvents:·pauses.count    avgt        382.000              #
SimpleBenchmark.publishSimpleEvents:·pauses.p0.00    avgt          1.976             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.50    avgt          4.184             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.90    avgt          7.387             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.95    avgt          8.518             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.99    avgt         11.092             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.999   avgt         14.172             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.9999  avgt         14.172             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p1.00    avgt         14.172             ms
```

~100M events/second feels about right for SPSC throughput. Don't ask me where I get that instinct from, though.

Palmr:
Since removing the modulo in favour of an `AtomicLong` I now get the following results on my `Intel(R) Core(TM) i5-6600K CPU @ 3.50GHz`:

```
Benchmark                                            Mode  Cnt   Score   Error  Units
SimpleBenchmark.publishSimpleEvents                  avgt    5   7.253 ± 0.167  ns/op
SimpleBenchmark.publishSimpleEvents:·pauses          avgt       16.843             ms
SimpleBenchmark.publishSimpleEvents:·pauses.avg      avgt       16.843             ms
SimpleBenchmark.publishSimpleEvents:·pauses.count    avgt        1.000              #
SimpleBenchmark.publishSimpleEvents:·pauses.p0.00    avgt       16.843             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.50    avgt       16.843             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.90    avgt       16.843             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.95    avgt       16.843             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.99    avgt       16.843             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.999   avgt       16.843             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p0.9999  avgt       16.843             ms
SimpleBenchmark.publishSimpleEvents:·pauses.p1.00    avgt       16.843             ms
```

grumpyjames:
Paired on this, but should probably try to keep a reasonable externally visible view of what we discussed.

**Thoughts**
This is not an ideal benchmark for either the disruptor or for array blocking queue. That's OK, because this commit's main aim is to help us start to use jmh in preference to our own performance test suites. Admittedly, I'm now wondering if we will discover on that journey that jmh is perhaps not what we want. To find that out, however, we're going to have to walk this path a bit further.

PR #315, opened 2020-11-07
Exclude currently intermittent tests

Palmr:
Add an `IntermittentTests` category and exclude tests categorised as such when running in CI until they get fixed.
Hopefully this should stop CI failing every other commit, but the tests are still fine to run locally rather than being ignored.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -191,3 +191,9 @@ class Version {
         "$major.$minor.$revision${stage ? '.' + stage : ''}${snapshot ? '-SNAPSHOT' : ''}"
     }
 }
+
+test {
+    useJUnit {
+        excludeCategories 'com.lmax.disruptor.IntermittentTests'
+    }
+}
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -191,3 +191,9 @@ class Version {
         "$major.$minor.$revision${stage ? '.' + stage : ''}${snapshot ? '-SNAPSHOT' : ''}"
     }
 }
+
+test {
```
eliquinox, line 195: Would it be better to create a separate Gradle task like `testNonIntermittent` for this purpose and point CI to run that task instead?

Palmr: Possibly, though as these tests are not ignored, they're still run locally if you want them to, so not sure adding a Gradle task for it would give us much?
I'm also hoping this category is short lived.

eliquinox (approved):

PR #316, opened 2020-11-08
Make use of lambdas, method references and implicit generics

eliquinox:
Refactor anonymous classes and explicit generics.

### DIFF
--- a/src/main/java/com/lmax/disruptor/dsl/ConsumerRepository.java
+++ b/src/main/java/com/lmax/disruptor/dsl/ConsumerRepository.java
@@ -62,22 +62,10 @@ class ConsumerRepository<T> implements Iterable<ConsumerInfo>
 
     public boolean hasBacklog(long cursor, boolean includeStopped)
     {
-        for (ConsumerInfo consumerInfo : consumerInfos)
-        {
-            if ((includeStopped || consumerInfo.isRunning()) && consumerInfo.isEndOfChain())
-            {
-                final Sequence[] sequences = consumerInfo.getSequences();
-                for (Sequence sequence : sequences)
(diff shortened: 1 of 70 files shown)


src/main/java/com/lmax/disruptor/dsl/ConsumerRepository.java
```diff
@@ -62,22 +62,10 @@ public void add(final WorkerPool<T> workerPool, final SequenceBarrier sequenceBa
...
-                    if (cursor > sequence.get())
-                    {
-                        return true;
-                    }
-                }
-            }
-        }
-
-        return false;
+        return consumerInfos.stream()
```
Palmr, line 65: This is not a change to use lambdas and implicit generics, to me this change is _more_ than that.

This code is only called on shutdown, but as one change out of many in this pull request this puts me on edge.

If this was a highly used function, what would the performance change be doing it the streams way vs the explicit loop way? Did you measure it? Is there even a benchmark covering it? etc.

While I don't think it matters much for `hasBacklog` here I'd still rather have this bit reverted for this pull request.

Palmr (changes requested):
Aside from that one comment made already, all seems alright.

src/main/java/com/lmax/disruptor/dsl/ConsumerRepository.java
eliquinox: Yes, if this was highly used I would not have delved into refactoring before understanding performance effects of such refactoring. But you are correct, it is best to keep the scientific hat on regardless. Change incoming.

Palmr (approved):

Issue #297, opened 2020-05-27
Markdown LMAX Documentation


emilypi:
Hi all, 

It would be great if the [docs](https://github.com/LMAX-Exchange/disruptor/blob/master/docs/Disruptor.docx) could be translated into markdown, so that they render well in browser. Would you mind a PR to that effect?

sea36:
We'd be happy to consider a pull request.

ronja-sigurdsdottir:
Great, will do @sea36

emilypi:
Great, will do thanks @sea36

Palmr:
Documentation had now been moved to ASCIIDoc(tor) rather than markdown. But that also renders will in a browser.
Hopefully that resolves this issue.

emilypi:
Ah, great. Thank you @Palmr! This was still on my radar, but I'm happy someone went ahead with it.

PR #317, opened 2020-11-09
Replace old docs with "living" ASCIIDoc(tor) documentation

swarren12:
The current method of documentation (Markdown documents + a MS Word document + JavaDoc + Github Wiki) is quite cumbersome:

* Markdown is nice and simple, but lacks more advanced features (e.g. being able to include snippets of source code directly from their origin files)
* Having the JavaDoc committed under `docs/` is just asking for it to be neglected.
* The Github Wiki appears to have several old pages (e.g. https://github.com/LMAX-Exchange/disruptor/wiki/Disruptor-Wizard, https://github.com/LMAX-Exchange/disruptor/wiki/Code-Example-Disruptor2x) and it is not clear (at least to me) whether these are still relevant given that we're now on v3.x

The changes in this PR will do the following:

* Replace the Markdown + MS Word + Github Wiki documentation with a single, unified version in ASCIIDoc
* Remove the static JavaDoc and instead generate this at build time, ensuring it is always up-to-date
* Add a new CI action that will regenerate the `gh-pages` branch from the ASCIIDoc + JavaDoc each time it runs

Example: 
- https://swarren12.github.io/disruptor/
- https://swarren12.github.io/disruptor/disruptor.html (Technical Paper)
- https://swarren12.github.io/disruptor/user-guide/ (User Guide)

Some things to consider: 

* The documentation (either the original sources or the version in this PR) probably needs to be reviewed as there are entire sections with no meaningful content (e.g. "Design and Implementation) - but that's outside of the scope of what I'm trying to achieve here!
* Should the "living" documentation be generated from `master` or from the latest release (/ should we have two versions)?


swarren12:
Not sure why the CI job is failing - possibly a permissions thing allowing the action to push to the branch?

Palmr (commented):
I see a removal of `docs/Disruptor.docx` documentation as well as the `docs/Disruptor-1.0.pdf`?
I can kind of understand removing the docx, assuming its contents had been transformed into asciidoc (which I don't think they have?) but the pdf should still be a resource linked to from the index?

I think the pdf should have stayed and the docx too, until it's fully replaced.

Palmr:
I think the `Generate GitHub Pages from ASCIIDoc` CI job needs an alternative for pull requests.
On upstream master branch it makes sense to run and deploy, on pull requests and we **really** don't want it blasting the `gh-pages` branch.
But it would be nice for it to build the pages to check the content is valid for pull requests and all branches.

Palmr:
> +1,572 −41,779

Excellent diff stats on the lines changed with this commit

swarren12:
> I see a removal of `docs/Disruptor.docx` documentation as well as the `docs/Disruptor-1.0.pdf`?
> I can kind of understand removing the docx, assuming its contents had been transformed into asciidoc (which I don't think they have?) but the pdf should still be a resource linked to from the index?
> 
> I think the pdf should have stayed and the docx too, until it's fully replaced.

Ah, I think I didn't read the content of these closely enough - I thought the `.docx` and the `.pdf` were the same (and thus both replaced by `disruptor.adoc`) but if they're different then I can add the PDF back in



> I think the `Generate GitHub Pages from ASCIIDoc` CI job needs an alternative for pull requests.
> On upstream master branch it makes sense to run and deploy, on pull requests and we **really** don't want it blasting the `gh-pages` branch.
> But it would be nice for it to build the pages to check the content is valid for pull requests and all branches.

Fair point - tbh I'm still not super familiar with the Github CI actions, so just copied an example! I'll take a look at what we can do. I'd guess having an action that builds the docs for all branches, but only pushing them for `master` should be easy enough.

Palmr:
>  I thought the .docx and the .pdf were the same (and thus both replaced by disruptor.adoc) but if they're different then I can add the PDF back in

You thought correctly and I was wrong. Keep them deleted.

swarren12:
I've updated the CI workflow to now always run the documentation generation on ever branch (to verify that it's valid) but only publish on `master`.

On a site note, is there any reason that the main Java build only happens for `master` or Pull Requests to `master`? Seems like that workflow (action? still not 100% clear on these!) should run on every branch?

Also, merging this PR probably closes issue #297 ?

Palmr:
> On a side note, is there any reason that the main Java build only happens for master or Pull Requests to master? Seems like that workflow (action? still not 100% clear on these!) should run on every branch?

Not really. Though moving to trunk-based development instead of all the branches should make this a moot point.
I wouldn't deny another pull request letting GitHub CI run on all branches however.

PR #322, opened 2020-11-13
Retrospectively change the meaning of halt on WorkProcessor

Palmr:
Our previous intermittency fix changed the meaning of `WorkProcessor::halt` and we should have updated the documentation to reflect that breaking change.


grumpyjames:
This looks fine, merge is imminent - I'll open a separate issue about the future of WorkerPool et al

PR #325, opened 2020-11-19
Restore Disruptor-1.0.pdf

swarren12:
Many places on the Internet link directly to this PDF, and updating all
of those links seems like a bit of a faff. It would be nice if in the
future we could generate the PDF from the `.adoc` as well, but for now
it's easier to just re-introduce the original version


Palmr:
> It would be nice if in the future we could generate the PDF from the .adoc as well

That would be excellent, if possible

Palmr (approved):

swarren12:
Original PDF restored: https://swarren12.github.io/disruptor/files/Disruptor-1.0.pdf

PR #328, opened 2020-11-20
Added JUnit 5 dependency

eliquinox:


Palmr (approved):
Looks like a nice small change. :+1:

PR #329, opened 2020-11-20
Update Gradle wrapper to v6.5

swarren12:
Update the Gradle wrapper to v6.5, and do the bare minimum* to get everything to still work.

*the bare minimum = remove the `osgi` plugin, because it no longer exists.

I'm not sure _why_ the OSGI plugin was required, although it does look as though someone wanted OSGI metadata in the past. It appears that the `biz.aQute.bnd.builder` plugin can re-enable the OSGI behaviour if required.


Palmr (changes requested):
> can re-enable the OSGI behaviour if required

Sounds good to me, there seems to be little documented reason for OSGI support.
Speaking of documenting things, perhaps a note in the `CHANGELOG` to say we're removing OSGI support?

Palmr (approved):

mattrpav:
OSGi support is used by Apache Karaf for supporting log4j2 async support using LMAX disruptor. The MANIFEST.MF metatdata is fairly minimal. Are you open to having it added back?

Palmr:
> OSGi support is used by Apache Karaf for supporting log4j2 async support using LMAX disruptor. The MANIFEST.MF metatdata is fairly minimal. Are you open to having it added back?

We're open to having it added back.
We could find no documented reason for it to exist, so it got removed. But if there's a valid use case we could add it back.
Happy to accept an issue/pull request, or we can look in to it.

mattrpav:
@Palmr thanks. I created #380

PR #341, opened 2020-12-29
Upload artifacts from GitHub Actions

Palmr:
We run a `gradle build` with GitHub Actions for CI but we were not exposing the reports generated by the build.

We also used to run the gradle build action on the master branch only, which meant any work on other branches was untested and could have gone all the way to merge request before flagging any issues.


swarren12:
LGTM

PR #342, opened 2020-12-29
JCStress test for Sequence

Palmr:
`Sequence` is a very important class to the Disruptor and it has methods like `set` and `setVolatile` which do subtly different things. It also makes general promises for thread safety/atomicity.

Given Sequence has existed for quite some time, this is less of an exercise in testing and more one of documenting existing behaviour.
There are comments on the methods of `Sequence` explaining the Store/Store and Store/Load rules of some of the methods, hopefully this test backs up those comments with some examples.

With this in place the move away from `UNSAFE` to `VarHandle` will be a little more of a proven non-breaking change.

### DIFF
--- a/.github/workflows/jcstress-sanity.yml
+++ b/.github/workflows/jcstress-sanity.yml
@@ -1,0 +1,26 @@
+name: JCStress Testing - commit-level sanity check
+
+on: [ push, pull_request ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ 11 ]
+    name: Java ${{ matrix.java }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      - name: Build with Gradle
+        run: ./gradlew jcstress -Pmode=sanity
+      - name: Archive artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: jcstress-report-sanity
+          path: build/reports/jcstress
--- a/.github/workflows/jcstress-stress.yml
+++ b/.github/workflows/jcstress-stress.yml
@@ -3,6 +3,7 @@ name: JCStress Testing - Thorough stress testing
 on:
   schedule:
     - cron: "0 0 * * WED"
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/jcstress-weekly-stress.yml
+++ b/.github/workflows/jcstress-weekly-stress.yml
@@ -1,0 +1,28 @@
+name: JCStress Testing - Thorough stress testing
+
+on:
+  schedule:
+    - cron: "0 0 * * WED"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ 11 ]
+    name: Java ${{ matrix.java }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up java
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+      - name: Build with Gradle
+        run: ./gradlew jcstress -Pmode=stress
+      - name: Archive artifacts
+        uses: actions/upload-artifact@v2
+        with:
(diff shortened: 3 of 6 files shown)


sea36:
Interesting tests. Forced me to go read the JCStress documentation!

In a couple of instances, you're using `J_Result` (for longs), but in others you use `L_Result` or `LL_Results` which have Object typed fields.  I think these should all be `J_Result` or `JJ_Result`.

I wonder if the CompareAndSet case would be better as:

```
    @JCStressTest
    @Outcome(id = {"true, false, 10", "false, true, 20"}, expect = ACCEPTABLE, desc = "Either updated.")
    @Outcome(expect = FORBIDDEN, desc = "Other cases are forbidden.")
    @State
    public static class CompareAndSet
    {
        Sequence sequence = new Sequence(0);

        @Actor
        public void actor1(ZZJ_Result r)
        {
            r.r1 = sequence.compareAndSet(0, 10);
        }

        @Actor
        public void actor2(ZZJ_Result r)
        {
            r.r2 = sequence.compareAndSet(0, 20);
        }

        @Arbiter
        public void arbiter(ZZJ_Result r)
        {
            r.r3 = sequence.get();
        }
    }
```

Palmr:
You're correct about the `L_result` types, I think `J_result` is probably the way they should be.

> I wonder if the CompareAndSet case would be better as

Yes, this is a good thing. I did attempt something like that myself initially but got in quite a muddle. This is very clean and simple though.

grumpyjames:
Well you're not forcing me into reading any documentation; I just run stuff to see what happens.

All of these results are from `OpenJDK 64-Bit Server VM Zulu11.37+17-CA (build 11.0.6+10-LTS, mixed mode)`

If I remove the volatile from `Value.value`, I get failures in `SetVolatileGuard` and `SetGuard` - that seems legitimate.

If I make the implementation of `setVolatile` `this.value = value`, nothing fails. That leads me to question the value of that `Unsafe` - is it faster than just setting the field, or am I missing something? Are we just making all the accesses via the `Unsafe` to make the pattern easier to spot.

If I make the implementation of `setVolatile` `UNSAFE.putOrderedLong(this, VALUE_OFFSET, value);`, I get failures from `SetVolatileDekker`. I google "Dekker" and wonder if you mean the Elizabethan dramatist or the 2006 Tirreno Adriatico winner. I stare at `SetVolatileDekker` for a while and convince myself that it tests something useful. OK, fine.

If I make the implementation of `set` `UNSAFE.putLong(this, VALUE_OFFSET, value);`, I don't see any failures. **This feels like we're missing a test**. Please don't ask me what the test looks like.

Unrelated: I can only find calls from test code to `incrementAndGet`. I wonder if that's a candidate for deletion in 4.x.

Palmr:
> I google "Dekker" and wonder if you mean the Elizabethan dramatist or the 2006 Tirreno Adriatico winner. 

@grumpyjames https://en.wikipedia.org/wiki/Dekker's_algorithm

Code shamelessly inspired by https://github.com/openjdk/jcstress/blob/master/jcstress-samples/src/main/java/org/openjdk/jcstress/samples/JMMSample_05_TotalOrder.java

sea36:
I agree the implementation of `setVolatile` could be simplified.  Looking at history, I think the `UNSAFE` implementation dates from the days when padding was achieved by holding the value in the middle of an array of `long[]`.  Not that padding is achieved by class hierarchy, and the field is declared `volatile long` it is redundant.  `getVolatile` used to do something similar, and that was changed to a simple `return value` a while ago.

lyncmi07:
I would also agree that using this UNSAFE stuff for the volatile field no longer looks neccessary, but given that we're planning on getting away from using UNSAFE anyway, I'm not sure it's worth getting to worked up over :smile: 

I'm happy with this change, would definitely make us much more confident about making changes to Sequence in future

lyncmi07 (approved):

grumpyjames:
I've been talking with @sea36 offline.

I still think we might be missing a test to show the value of `putOrdered` vs. `put`, but I think we'll need a non-x86 arch to test it on (something ARM based - anyone on a new fancy macbook? No? How about a r-pi?).

I think we should merge this.

grumpyjames (approved):

lyncmi07:
> I've been talking with @sea36 offline.
> 
> I still think we might be missing a test to show the value of `putOrdered` vs. `put`, but I think we'll need a non-x86 arch to test it on (something ARM based - anyone on a new fancy macbook? No? How about a r-pi?).
> 
> I think we should merge this.

I'm happy to expense a new macbook :D

PR #346, opened 2021-01-12
CodeQL workflow

fillreject:


.github/workflows/codeql-analysis.yml
```diff
@@ -59,9 +59,16 @@ jobs:
...
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+    
+    - name: Build with Gradle
+      run: ./gradlew build
+
```
Palmr, line 72: Might be nice to remove all the commented out bits from the default file and leave a comment explaining why we had to build manually instead of using the autobuild step

README.md
```diff
@@ -1,6 +1,7 @@
 ## LMAX Disruptor
 
 ![Java CI with Gradle](https://github.com/LMAX-Exchange/disruptor/workflows/Java%20CI%20with%20Gradle/badge.svg)
+[![CodeQL](https://github.com/fillreject/disruptor/workflows/CodeQL/badge.svg)](https://github.com/fillreject/disruptor/actions?query=workflow%3ACodeQL)
```
Palmr, line 4: Not sure we can take this, linked to the @fillreject fork.
Might be better to have the PR without changes to this badge and do the change direct on the repo after

Palmr (changes requested):
Tool seems to work which is good, but the changes to the badge links need to be reverted before we accept this.

README.md
fillreject: I removed the commits with this change.

Palmr (approved):

PR #347, opened 2021-01-13
remove the obsolete TODO comment

beyondacm:
##### SUMMARY
Remove the obsolete TODO comment. 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

src/main/java/com/lmax/disruptor/MultiProducerSequencer.java (line:136)

##### ADDITIONAL INFORMATION

The pending task of the TODO comment has already performed in the earlier version, but the TODO comment was not removed accordingly at the moment. Please check the following commit: 
https://github.com/LMAX-Exchange/disruptor/commit/85905a0f48db8a0997eb1a5c0bf5ec878d399345

### DIFF
--- a/src/main/java/com/lmax/disruptor/MultiProducerSequencer.java
+++ b/src/main/java/com/lmax/disruptor/MultiProducerSequencer.java
@@ -133,7 +133,7 @@ public final class MultiProducerSequencer extends AbstractSequencer
 
                 if (wrapPoint > gatingSequence)
                 {
-                    LockSupport.parkNanos(1); // TODO, should we spin based on the wait strategy?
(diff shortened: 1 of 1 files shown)


Palmr:
That pending TODO has not been performed in an earlier version.
It is still using `LockSupport.parkNanos(1L);` to spin and not using the waitStrategy.

The call to `waitStrategy` in that loop, as referenced by 85905a0 in your PR, is not related to the TODO you have removed.

This TODO is still an open question, so this PR cannot be merged.

Your change also caused CI to fail the checkstyle verification. Removing the comment only left the trailing space between the remaining line of code and where the comment was.

Issue #231, opened 2018-09-09
Padding should use 2 cache lines to pad


franz1981:
HI!!!

given that prefetchers use 2 cache lines, `Sequence` (or similar padded classes) shouldn't be instead:
```
    static class LhsPadding
    {
        protected long p1, p2, p3, p4, p5, p6, p7;
        protected long p8, p9, p10, p11, p12, p13, p14, p15;
    }

    static class Value extends LhsPadding
    {
        protected volatile long value;
    }

    static class RhsPadding extends Value
    {
        protected long p1, p2, p3, p4, p5, p6, p7;
        protected long p8, p9, p10, p11, p12, p13, p14, p15;
    }
```

To allow with both compressed oop or not:
```
RhsPadding object internals:
 OFFSET  SIZE   TYPE DESCRIPTION                               VALUE
      0    12        (object header)                           N/A
     12     4        (alignment/padding gap)                  
     16     8   long LhsPadding.p1                             N/A
     24     8   long LhsPadding.p2                             N/A
     32     8   long LhsPadding.p3                             N/A
     40     8   long LhsPadding.p4                             N/A
     48     8   long LhsPadding.p5                             N/A
     56     8   long LhsPadding.p6                             N/A
     64     8   long LhsPadding.p7                             N/A
     72     8   long LhsPadding.p8                             N/A
     80     8   long LhsPadding.p9                             N/A
     88     8   long LhsPadding.p10                            N/A
     96     8   long LhsPadding.p11                            N/A
    104     8   long LhsPadding.p12                            N/A
    112     8   long LhsPadding.p13                            N/A
    120     8   long LhsPadding.p14                            N/A
    128     8   long LhsPadding.p15                            N/A
    136     8   long Value.value                               N/A
    144     8   long RhsPadding.p1                             N/A
    152     8   long RhsPadding.p2                             N/A
    160     8   long RhsPadding.p3                             N/A
    168     8   long RhsPadding.p4                             N/A
    176     8   long RhsPadding.p5                             N/A
    184     8   long RhsPadding.p6                             N/A
    192     8   long RhsPadding.p7                             N/A
    200     8   long RhsPadding.p8                             N/A
    208     8   long RhsPadding.p9                             N/A
    216     8   long RhsPadding.p10                            N/A
    224     8   long RhsPadding.p11                            N/A
    232     8   long RhsPadding.p12                            N/A
    240     8   long RhsPadding.p13                            N/A
[...]

PR #352, opened 2021-02-01
Move false-sharing padding to be in bytes rather than longs

langera:
To add to 4.0 work..

This is to make it work with JDK 15 memory layout changes.
see: https://shipilev.net/jvm/objects-inside-out/#_observation_hierarchy_tower_padding_trick_collapse_in_jdk_15

Should be harmless in earlier JDKs.

The changes save a bit of memory (up to 16 bytes on RingBuffer) by being able to fit bytes into internal JVM padding after object header etc. (You can see it with the JavaObjectLayout tool).
As far as false-sharing is concerned, I think that the Sequence changes are somewhat academic since we're trying to protect a single long from false-sharing and the new approach will not move it regardless. Still, it makes the code slightly less reliant on implementation quirks. 
With the ring buffer fields, this change will still prevent false sharing if the memory ordering changes by the layout.

### DIFF
--- a/src/jmh/java/com/lmax/disruptor/RingBufferFalseSharingBenchmark.java
+++ b/src/jmh/java/com/lmax/disruptor/RingBufferFalseSharingBenchmark.java
@@ -1,0 +1,309 @@
+package com.lmax.disruptor;
+
+import com.lmax.disruptor.util.SimpleEvent;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Group;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+import java.util.concurrent.TimeUnit;
+
+/*
+ * Based on false-sharing benchmark in open JDK
+ * @see https://github.com/openjdk/jmh/blob/master/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_22_FalseSharing.java
+ *
+ */
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Fork(5)
+public class RingBufferFalseSharingBenchmark
+{
+    /*
+     * We take advantage of the inheritance trick used in RingBuffer
+     * to create an object without the padding that occur after the fields.
+     *
+     * Java object layout using JDK15:
+     * com.lmax.disruptor.RingBufferFalseSharingBenchmark$HalfPaddedRingBufferWithNoisyNeighbour object internals:
+ OFFSET  SIZE                           TYPE DESCRIPTION                                        VALUE
+      0    12                                (object header)                                    N/A
+     12     1                           byte RingBufferPad.p77                                  N/A
+     13     1                           byte RingBufferPad.p11                                  N/A
+     14     1                           byte RingBufferPad.p12                                  N/A
+     15     1                           byte RingBufferPad.p13                                  N/A
+     16     1                           byte RingBufferPad.p14                                  N/A
+     17     1                           byte RingBufferPad.p15                                  N/A
+     18     1                           byte RingBufferPad.p16                                  N/A
+     19     1                           byte RingBufferPad.p17                                  N/A
+     20     1                           byte RingBufferPad.p20                                  N/A
+     21     1                           byte RingBufferPad.p21                                  N/A
+     22     1                           byte RingBufferPad.p22                                  N/A
+     23     1                           byte RingBufferPad.p23                                  N/A
+     24     1                           byte RingBufferPad.p24                                  N/A
+     25     1                           byte RingBufferPad.p25                                  N/A
+     26     1                           byte RingBufferPad.p26                                  N/A
+     27     1                           byte RingBufferPad.p27                                  N/A
+     28     1                           byte RingBufferPad.p30                                  N/A
+     29     1                           byte RingBufferPad.p31                                  N/A
+     30     1                           byte RingBufferPad.p32                                  N/A
+     31     1                           byte RingBufferPad.p33                                  N/A
+     32     1                           byte RingBufferPad.p34                                  N/A
+     33     1                           byte RingBufferPad.p35                                  N/A
+     34     1                           byte RingBufferPad.p36                                  N/A
+     35     1                           byte RingBufferPad.p37                                  N/A
+     36     1                           byte RingBufferPad.p40                                  N/A
+     37     1                           byte RingBufferPad.p41                                  N/A
+     38     1                           byte RingBufferPad.p42                                  N/A
+     39     1                           byte RingBufferPad.p43                                  N/A
+     40     1                           byte RingBufferPad.p44                                  N/A
+     41     1                           byte RingBufferPad.p45                                  N/A
+     42     1                           byte RingBufferPad.p46                                  N/A
+     43     1                           byte RingBufferPad.p47                                  N/A
+     44     1                           byte RingBufferPad.p50                                  N/A
+     45     1                           byte RingBufferPad.p51                                  N/A
+     46     1                           byte RingBufferPad.p52                                  N/A
+     47     1                           byte RingBufferPad.p53                                  N/A
+     48     1                           byte RingBufferPad.p54                                  N/A
+     49     1                           byte RingBufferPad.p55                                  N/A
+     50     1                           byte RingBufferPad.p56                                  N/A
+     51     1                           byte RingBufferPad.p57                                  N/A
+     52     1                           byte RingBufferPad.p60                                  N/A
+     53     1                           byte RingBufferPad.p61                                  N/A
+     54     1                           byte RingBufferPad.p62                                  N/A
+     55     1                           byte RingBufferPad.p63                                  N/A
+     56     1                           byte RingBufferPad.p64                                  N/A
+     57     1                           byte RingBufferPad.p65                                  N/A
+     58     1                           byte RingBufferPad.p66                                  N/A
+     59     1                           byte RingBufferPad.p67                                  N/A
+     60     1                           byte RingBufferPad.p70                                  N/A
+     61     1                           byte RingBufferPad.p71                                  N/A
+     62     1                           byte RingBufferPad.p72                                  N/A
+     63     1                           byte RingBufferPad.p73                                  N/A
+     64     1                           byte RingBufferPad.p74                                  N/A
+     65     1                           byte RingBufferPad.p75                                  N/A
+     66     1                           byte RingBufferPad.p76                                  N/A
+     67     1                           byte RingBufferPad.p10                                  N/A
+     68     4                            int RingBufferFields.bufferSize                        N/A
+     72     8                           long RingBufferFields.indexMask                         N/A
+     80     4             java.lang.Object[] RingBufferFields.entries                           N/A
+     84     4   com.lmax.disruptor.Sequencer RingBufferFields.sequencer                         N/A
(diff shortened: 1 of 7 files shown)


Palmr:
For reference, some of this was already discussed in #231 
Though that was `Sequence` only which as you rightly point out if a bit of an academic change to make. We didn't look in to `RingBuffer` and its padding.

Any chance you took a look at the layout in 11/15 with jol, or created any jmh benchmarks proving this change doesn't degrade performance (and ideally when run on jdk15 gives better performance)?

I'm trying to make sure any changes made for 4.0 have some kind of benchmark/proof backing up their performance claims so future changes are easier to make and reason about past decisions.

langera:
I understand your approach. 
But I'm not aware of any way to design a test to specifically produce false sharing as it is impossible to control where exactly the JVM will decide to split the memory into blocks. We can only prevent it, not create it.
What I do expect is to see in other perf tests that with padding (already there) they behave more consistently, that's all.
But that's all to do with the padding idea in general (and that's already there).

As discussed, the difference between the padding that is already there and the new "byte padding" in Sequence is nil in terms of performance. 
I'll check the exact layout of RingBuffer on JDK11, JDK15 and add it to this PR (hopefully tomorrow).

Because the padding is already there, my feeling was that regardless, its better to follow the recommended padding pattern rather than the old one which is documented as not working in certain scenarios, but I'll leave it up to you to decide.

Palmr:
Yes, I was thinking we have some JMH benchmarks on `Sequence` now which should start going slower on jdk15 if the layout was non-optimal and we started false-sharing.
I wonder if we could do the same with `RingBuffer` and friends? It won't be pass-fail, but a fine-grained JMH benchmark showing the ops/us that we cna run on jdk11 and jdk15 may show some slowdown.

If we can't have some kind of benchmark showing something like a lack of padding (or in this case, a non-optimal layout) causing a slow down due to false sharing, does the padding really do anything?

langera:
RingBuffer object layout:

**JDK 11**

```
com.lmax.disruptor.RingBuffer object internals:
 OFFSET  SIZE                           TYPE DESCRIPTION                               VALUE
      0    12                                (object header)                           N/A
     12     4                                (alignment/padding gap)
     16     8                           long RingBufferPad.p1                          N/A
     24     8                           long RingBufferPad.p2                          N/A
     32     8                           long RingBufferPad.p3                          N/A
     40     8                           long RingBufferPad.p4                          N/A
     48     8                           long RingBufferPad.p5                          N/A
     56     8                           long RingBufferPad.p6                          N/A
     64     8                           long RingBufferPad.p7                          N/A
     72     8                           long RingBufferFields.indexMask                N/A
     80     4                            int RingBufferFields.bufferSize               N/A
     84     4             java.lang.Object[] RingBufferFields.entries                  N/A
     88     4   com.lmax.disruptor.Sequencer RingBufferFields.sequencer                N/A
     92     4                                (alignment/padding gap)
     96     8                           long RingBuffer.p1                             N/A
    104     8                           long RingBuffer.p2                             N/A
    112     8                           long RingBuffer.p3                             N/A
    120     8                           long RingBuffer.p4                             N/A
    128     8                           long RingBuffer.p5                             N/A
    136     8                           long RingBuffer.p6                             N/A
    144     8                           long RingBuffer.p7                             N/A
Instance size: 152 bytes
Space losses: 8 bytes internal + 0 bytes external = 8 bytes total
```

**JDK 15**

```
com.lmax.disruptor.RingBuffer object internals:
 OFFSET  SIZE                           TYPE DESCRIPTION                               VALUE
      0    12                                (object header)                           N/A
     12     4                            int RingBufferFields.bufferSize               N/A
     16     8                           long RingBufferPad.p1                          N/A
     24     8                           long RingBufferPad.p2                          N/A
     32     8                           long RingBufferPad.p3                          N/A
     40     8                           long RingBufferPad.p4                          N/A
     48     8                           long RingBufferPad.p5                          N/A
     56     8                           long RingBufferPad.p6                          N/A
     64     8                           long RingBufferPad.p7                          N/A
     72     8                           long RingBufferFields.indexMask                N/A
     80     4             java.lang.Object[] RingBufferFields.entries                  N/A
     84     4   com.lmax.disruptor.Sequencer RingBufferFields.sequencer                N/A
     88     8                           long RingBuffer.p1                             N/A
     96     8                           long RingBuffer.p2                             N/A
    104     8                           long RingBuffer.p3                             N/A
    112     8                           long RingBuffer.p4                             N/A
    120     8                           long RingBuffer.p5                             N/A
    128     8                           long RingBuffer.p6                             N/A
    136     8                           long RingBuffer.p7                             N/A
Instance size: 144 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

Note `bufferSize` was moved to the gap after the object header and is no longer "protected" by the padding.

langera:
**JDK 15 on PR fork**
```
com.lmax.disruptor.RingBuffer object internals:
 OFFSET  SIZE                           TYPE DESCRIPTION                               VALUE
      0    12                                (object header)                           N/A
     12     1                           byte RingBufferPad.p77                         N/A
     13     1                           byte RingBufferPad.p11                         N/A
     14     1                           byte RingBufferPad.p12                         N/A
     15     1                           byte RingBufferPad.p13                         N/A
     16     1                           byte RingBufferPad.p14                         N/A
     17     1                           byte RingBufferPad.p15                         N/A
     18     1                           byte RingBufferPad.p16                         N/A
     19     1                           byte RingBufferPad.p17                         N/A
     20     1                           byte RingBufferPad.p20                         N/A
     21     1                           byte RingBufferPad.p21                         N/A
     22     1                           byte RingBufferPad.p22                         N/A
     23     1                           byte RingBufferPad.p23                         N/A
     24     1                           byte RingBufferPad.p24                         N/A
     25     1                           byte RingBufferPad.p25                         N/A
     26     1                           byte RingBufferPad.p26                         N/A
     27     1                           byte RingBufferPad.p27                         N/A
     28     1                           byte RingBufferPad.p30                         N/A
     29     1                           byte RingBufferPad.p31                         N/A
     30     1                           byte RingBufferPad.p32                         N/A
     31     1                           byte RingBufferPad.p33                         N/A
     32     1                           byte RingBufferPad.p34                         N/A
     33     1                           byte RingBufferPad.p35                         N/A
     34     1                           byte RingBufferPad.p36                         N/A
     35     1                           byte RingBufferPad.p37                         N/A
     36     1                           byte RingBufferPad.p40                         N/A
     37     1                           byte RingBufferPad.p41                         N/A
     38     1                           byte RingBufferPad.p42                         N/A
     39     1                           byte RingBufferPad.p43                         N/A
     40     1                           byte RingBufferPad.p44                         N/A
     41     1                           byte RingBufferPad.p45                         N/A
     42     1                           byte RingBufferPad.p46                         N/A
     43     1                           byte RingBufferPad.p47                         N/A
     44     1                           byte RingBufferPad.p50                         N/A
     45     1                           byte RingBufferPad.p51                         N/A
     46     1                           byte RingBufferPad.p52                         N/A
     47     1                           byte RingBufferPad.p53                         N/A
     48     1                           byte RingBufferPad.p54                         N/A
     49     1                           byte RingBufferPad.p55                         N/A
     50     1                           byte RingBufferPad.p56                         N/A
     51     1                           byte RingBufferPad.p57                         N/A
     52     1                           byte RingBufferPad.p60                         N/A
     53     1                           byte RingBufferPad.p61                         N/A
     54     1                           byte RingBufferPad.p62                         N/A
     55     1                           byte RingBufferPad.p63                         N/A
     56     1                           byte RingBufferPad.p64                         N/A
     57     1                           byte RingBufferPad.p65                         N/A
     58     1                           byte RingBufferPad.p66                         N/A
     59     1                           byte RingBufferPad.p67                         N/A
     60     1                           byte RingBufferPad.p70                         N/A
     61     1                           byte RingBufferPad.p71                         N/A
     62     1                           byte RingBufferPad.p72                         N/A
     63     1                           byte RingBufferPad.p73                         N/A
     64     1                           byte RingBufferPad.p74                         N/A
     65     1                           byte RingBufferPad.p75                         N/A
     66     1                           byte RingBufferPad.p76                         N/A
     67     1                           byte RingBufferPad.p10                         N/A
     68     4                            int RingBufferFields.bufferSize               N/A
     72     8                           long RingBufferFields.indexMask                N/A
     80     4             java.lang.Object[] RingBufferFields.entries                  N/A
     84     4   com.lmax.disruptor.Sequencer RingBufferFields.sequencer                N/A
     88     1                           byte RingBuffer.p77                            N/A
     89     1                           byte RingBuffer.p11                            N/A
     90     1                           byte RingBuffer.p12                            N/A
     91     1                           byte RingBuffer.p13                            N/A
     92     1                           byte RingBuffer.p14                            N/A
     93     1                           byte RingBuffer.p15                            N/A
     94     1                           byte RingBuffer.p16                            N/A
     95     1                           byte RingBuffer.p17                            N/A
     96     1                           byte RingBuffer.p20                            N/A
     97     1                           byte RingBuffer.p21                            N/A
     98     1                           byte RingBuffer.p22                            N/A
     99     1                           byte RingBuffer.p23                            N/A
    100     1                           byte RingBuffer.p24                            N/A
    101     1                           byte RingBuffer.p25                            N/A
    102     1                           byte RingBuffer.p26                            N/A
    103     1                           byte RingBuffer.p27                            N/A
    104     1                           byte RingBuffer.p30                            N/A
    105     1                           byte RingBuffer.p31                            N/A
    106     1                           byte RingBuffer.p32                            N/A
    107     1                           byte RingBuffer.p33                            N/A
    108     1                           byte RingBuffer.p34                            N/A
    109     1                           byte RingBuffer.p35                            N/A
    110     1                           byte RingBuffer.p36                            N/A
    111     1                           byte RingBuffer.p37                            N/A
    112     1                           byte RingBuffer.p40                            N/A
    113     1                           byte RingBuffer.p41                            N/A
    114     1                           byte RingBuffer.p42                            N/A
    115     1                           byte RingBuffer.p43                            N/A
    116     1                           byte RingBuffer.p44                            N/A
    117     1                           byte RingBuffer.p45                            N/A
    118     1                           byte RingBuffer.p46                            N/A
    119     1                           byte RingBuffer.p47                            N/A
    120     1                           byte RingBuffer.p50                            N/A
    121     1                           byte RingBuffer.p51                            N/A
    122     1                           byte RingBuffer.p52                            N/A
    123     1                           byte RingBuffer.p53                            N/A
    124     1                           byte RingBuffer.p54                            N/A
    125     1                           byte RingBuffer.p55                            N/A
    126     1                           byte RingBuffer.p56                            N/A
    127     1                           byte RingBuffer.p57                            N/A
    128     1                           byte RingBuffer.p60                            N/A
    129     1                           byte RingBuffer.p61                            N/A
    130     1                           byte RingBuffer.p62                            N/A
    131     1                           byte RingBuffer.p63                            N/A
    132     1                           byte RingBuffer.p64                            N/A
    133     1                           byte RingBuffer.p65                            N/A
    134     1                           byte RingBuffer.p66                            N/A
    135     1                           byte RingBuffer.p67                            N/A
    136     1                           byte RingBuffer.p70                            N/A
    137     1                           byte RingBuffer.p71                            N/A
    138     1                           byte RingBuffer.p72                            N/A
    139     1                           byte RingBuffer.p73                            N/A
    140     1                           byte RingBuffer.p74                            N/A
    141     1                           byte RingBuffer.p75                            N/A
    142     1                           byte RingBuffer.p76                            N/A
    143     1                           byte RingBuffer.p10                            N/A
Instance size: 144 bytes
Space losses: 0 bytes internal + 0 bytes external = 0 bytes total
```

langera:
There are two points here:
1. The object layout of `RingBuffer` does change in JDK15.  
2. Being data led is the right approach. The problem here is getting hold of the data. To be able to reproduce false-sharing you need to control internal JVM (possibly random) decisions. Unless you write your own JVM and can control what goes into what memory block, seeing the effect can only be a general observation over time of more consistent results (not necessarily better performance). There are very few of these extreme cases where you need to either trust the recommended approach by JVM writers and follow it or just ignore it.

Palmr:
I get that this change is more about changing that padding to satisfy layout changes in jdk15, where writing a test for that case would be interesting to say the least.
But as you were touching the padding, which is there for false-sharing avoidance purposes I'd hoped to tempt you in to writing some JMH benchmarks for a padded/unpadded version of RingBuffer and friends, similar to https://github.com/openjdk/jmh/blob/master/jmh-samples/src/main/java/org/openjdk/jmh/samples/JMHSample_22_FalseSharing.java where it's quite easy to see the effects of false-sharing and the benefits of padding.

Such benchmarks aren't really required for me to see the benefits of the core of this change, which is the JDK15 layout changes. And you've nicely documented the reason for this change now with the JOL output in the PR.

I shall leave this PR open until tomorrow afternoon in case you feel the suitably tempted in to adding a false-sharing-demonstrating benchmark.
If not swayed, the changes are fine as they are and I'll merge them.

Palmr (approved):

langera:
I can take a stab at writing something like the benchmark you referenced, I just need to find the time :) 
I'll try to do something this coming week.

langera:
I wrote a benchmark that tries to check false-sharing padding performance impact while still not completely faking the RingBuffer object.
I used the fact we are using inheritance for the padding and so we can stop somewhere in the middle and lose half of the padding as a result. I hope this makes sense.

Palmr:
Looks good, I get similar results here showing a slight improvement on the fully padded version.
Thank you for taking the time and adding a benchmark for us.

PR #355, opened 2021-02-28
Increase MultiProducerSequencer performance on getting next sequence

bas-k:
The MultiProducerSequencer currently has contention problems when many threads try to get the next sequence. It is not necessary to loop trying to compareAndSet the new sequence number. The producer can allocate sequence numbers upfront as long as it does not return when the consumer has not consumed the slot yet. The up-front allocation is valid because the call to getHighestPublisedSequence is the one that will tell the consumer that a slot has been published by the publisher.

### DIFF
--- a/src/main/java/com/lmax/disruptor/MultiProducerSequencer.java
+++ b/src/main/java/com/lmax/disruptor/MultiProducerSequencer.java
@@ -116,37 +116,24 @@ public final class MultiProducerSequencer extends AbstractSequencer
             throw new IllegalArgumentException("n must be > 0");
         }
 
-        long current;
-        long next;
+        long current = cursor.getAndAdd(n);
 
-        do
+        long nextSequence = current + n;
+        long wrapPoint = nextSequence - bufferSize;
+        long cachedGatingSequence = gatingSequenceCache.get();
+
+        if (wrapPoint > cachedGatingSequence || cachedGatingSequence > current)
         {
-            current = cursor.get();
-            next = current + n;
-
-            long wrapPoint = next - bufferSize;
-            long cachedGatingSequence = gatingSequenceCache.get();
-
-            if (wrapPoint > cachedGatingSequence || cachedGatingSequence > current)
+            long gatingSequence;
+            while (wrapPoint > (gatingSequence = Util.getMinimumSequence(gatingSequences, current)))
             {
-                long gatingSequence = Util.getMinimumSequence(gatingSequences, current);
-
-                if (wrapPoint > gatingSequence)
-                {
-                    LockSupport.parkNanos(1); // TODO, should we spin based on the wait strategy?
-                    continue;
-                }
-
-                gatingSequenceCache.set(gatingSequence);
+                LockSupport.parkNanos(1L); // TODO, should we spin based on the wait strategy?
             }
-            else if (cursor.compareAndSet(current, next))
-            {
-                break;
-            }
+
+            gatingSequenceCache.set(gatingSequence);
         }
-        while (true);
 
-        return next;
+        return nextSequence;
     }
 
     /**
--- a/src/main/java/com/lmax/disruptor/Sequence.java
+++ b/src/main/java/com/lmax/disruptor/Sequence.java
@@ -103,6 +103,17 @@ public class Sequence extends RhsPadding
     }
 
     /**
+     * Perform an atomic getAndAdd operation on the sequence.
+     *
+     * @param increment The value to add to the sequence.
+     * @return the value before increment
+     */
+    public long getAndAdd(final long increment)
+    {
+        return UNSAFE.getAndAddLong(this, VALUE_OFFSET, increment);
+    }
+
+    /**
      * Performs a volatile write of this sequence.  The intent is
      * a Store/Store barrier between this write and any previous
      * write and a Store/Load barrier between this write and any
--- a/src/test/java/com/lmax/disruptor/RingBufferTest.java
+++ b/src/test/java/com/lmax/disruptor/RingBufferTest.java
@@ -190,7 +190,8 @@ public class RingBufferTest
         thread.start();
 
         latch.await();
-        assertThat(Long.valueOf(buffer2.getCursor()), is(Long.valueOf(ringBufferSize - 1)));
+        assertThat(Long.valueOf(buffer2.getCursor()), is(Long.valueOf(ringBufferSize)));
+        assertFalse(buffer2.isPublished(buffer2.getCursor()));
         assertFalse(publisherComplete.get());
 
         processor.run();
--- a/src/test/java/com/lmax/disruptor/SequencerTest.java
+++ b/src/test/java/com/lmax/disruptor/SequencerTest.java
@@ -93,7 +93,9 @@ public class SequencerTest
         final CountDownLatch doneLatch = new CountDownLatch(1);
 
         final long expectedFullSequence = Sequencer.INITIAL_CURSOR_VALUE + sequencer.getBufferSize();
-        assertThat(sequencer.getCursor(), is(expectedFullSequence));
+        assertThat(
+            sequencer.getHighestPublishedSequence(Sequencer.INITIAL_CURSOR_VALUE + 1, sequencer.getCursor()),
+            is(expectedFullSequence));
 
         executor.submit(
             new Runnable()
@@ -111,12 +113,14 @@ public class SequencerTest
             });
 
         waitingLatch.await();
-        assertThat(sequencer.getCursor(), is(expectedFullSequence));
+        assertThat(
+            sequencer.getHighestPublishedSequence(expectedFullSequence, sequencer.getCursor()),
+            is(expectedFullSequence));
 
         gatingSequence.set(Sequencer.INITIAL_CURSOR_VALUE + 1L);
 
         doneLatch.await();
-        assertThat(sequencer.getCursor(), is(expectedFullSequence + 1L));
+        assertThat(sequencer.getHighestPublishedSequence(expectedFullSequence, sequencer.getCursor()), is(expectedFullSequence + 1L));
     }
 
     @Test(expected = InsufficientCapacityException.class)


src/main/java/com/lmax/disruptor/MultiProducerSequencer.java
```diff
@@ -116,37 +116,24 @@ public long next(int n)
...
-            if (wrapPoint > cachedGatingSequence || cachedGatingSequence > current)
-            {
-                long gatingSequence = Util.getMinimumSequence(gatingSequences, current);
-
-                if (wrapPoint > gatingSequence)
-                {
-                    LockSupport.parkNanos(1); // TODO, should we spin based on the wait strategy?
-                    continue;
-                }
+        long nextSequence = current + n;
```
MendelMonteiro, line 121: Aren't you adding `n` twice here, once on this line and once on 119 ?

bas-k: The call `long nextSequence = cursor.getAndAdd(n)` on line 119 will return the current value before adding n. So the sequence to return should be `current + n`

MendelMonteiro: 👍

ocoanet:
I like the idea to remove the CAS loop in `MultiProducerSequencer`. It should significantly improve the sequencer performance. Did you run the tests with your change? Can you post the results?

Yet, the change will generate situations in which the `cursor` value will be out-of sync with `gatingSequences`. These situations will be very short most of the time but can last longer when the ring buffer is full. I think the impact of the out-of-sync state should be thoroughly analyzed. For example, I believe that `remainingCapacity()` could return negative values.

Also I think that code depending on `RingBuffer.getCursor()` could be broken in subtle ways.

I wonder if it is possible to keep your change but also update other methods like `remainingCapacity` and `getCursor` to make sure they behave as if the cursor was not incremented above the `gatingSequences` limit?

bas-k:
Here is some test data (not gathered using jmh). This was done with a producer that published an event as fast as possible and a consumer that just received events and did nothing further. The machine I tested this on had 48 cores.
The difference in performance gets bigger with more producer threads (as expected).

**Original MultiproducerSequencer**
run(ms) | queueSize | Producer threads | Number of events
-- | -- | -- | --
10000 | 4.194.304 | 2 | 38.927.860
10000 | 4.194.304 | 2 | 40.581.745
10000 | 4.194.304 | 2 | 38.106.458
10000 | 4.194.304 | 2 | 37.975.908
10000 | 4.194.304 | 2 | 34.863.021
10000 | 4.194.304 | 2 | 37.786.142
10000 | 4.194.304 | 32 | 9.516.499
10000 | 4.194.304 | 32 | 9.549.365
10000 | 4.194.304 | 32 | 8.655.449
10000 | 4.194.304 | 32 | 8.568.923
10000 | 4.194.304 | 32 | 8.524.075
10000 | 4.194.304 | 32 | 8.466.991

**Patched MultiproducerSequencer**
run(ms) | queueSize | Producer threads | Number of events
-- | -- | -- | --
10000 | 4.194.304 | 2 | 50.975.107
10000 | 4.194.304 | 2 | 51.617.768
10000 | 4.194.304 | 2 | 48.881.355
10000 | 4.194.304 | 2 | 48.692.707
10000 | 4.194.304 | 2 | 45.395.575
10000 | 4.194.304 | 2 | 42.117.900
10000 |  8.388.608 | 32 | 139.026.852
10000 |  8.388.608 | 32 | 140.893.555
10000 |  8.388.608 | 32 | 143.817.425
10000 |  8.388.608 | 32 | 143.372.026
10000 |  8.388.608 | 32 | 144.402.274
10000 |  8.388.608 | 32 | 141.399.396

`remainingCapacity()` can already return negative values with the current code. There is some time between the call to get the consumed sequence and the produced sequence.

>     long consumed = Util.getMinimumSequence(gatingSequences, cursor.get());
>     long produced = cursor.get();
>     return getBufferSize() - (produced - consumed);

In this small window, the cursor could be updated by some other thread requesting the next sequence, which would mean that the returned capacity is negative. I agree that this change will make a negative capacity more likely.

The value of `getCursor()` is always valid, because there is some producer wanting to put data into the queue. The problem is that you should not rely on the value of the cursor by itself. There should always be a call to `getHighestPublishedSequence()` or `isAvailable()` to check if the data has actually been put into the queue.

ocoanet:
I will try to recap the impacts.

Before:
- `getCursor` is unsafe: it can return an unpublished sequence
- `remainingCapacity` can return negative values (quite unlikely)
- the Disruptor gets broken if a thread crashes/hangs after claiming a sequence and before publishing it (very short time window)

After:
- `getCursor` is unsafe: it can return an unpublished sequence **for an event that is potentially still used by dependant processors**
- `remainingCapacity` can return negative values (**very likely when the ring buffer is full**)
- the Disruptor gets broken if a thread crashes/hangs **after incrementing the cursor and before publishing the sequence (possibly large time window when the ring buffer is full)**

Your change does not seem to affect `hasAvailableCapacity` or `getMinimumSequence`. The impact on `getCursor` is less clear for me (especially in `SequenceGroups.addSequences`) but this method was already dangerous before anyway.

I do not think that the performance of `getCursor` is critical, so it could be interesting to improve it by returning the last published sequence, or to add an alternative `getPublishedCursor` method. This could be the object of another PR.

Am I missing something important here?

It would be nice if we could get some feedback from the mainainters. The performance improvements of your change are just too good to be ignored 🙂

Palmr:
Maintainer here to provide some feedback :)

As far as I can see this is mostly a re-shuffling of the existing code, and the rough edges around methods such as `getCursor` were there before and remain there.
This change appears to move the CAS loop from one that spanned most of logic in `MultiProducerSequencer::next` to one small CAS loop when adding to the cursor initially, and a non-thread-contentious loop for gating.

In terms of edge cases, as I see it all cases existed before and now may be more likely but nothing new has been introduced.
Therefore the performance improvements are indeed seem too good to be ignored.

Questions and suggestions:

 - I see this change was made atop the 3.4.2 version commit. Is there any reason not to make the change on latest master, to go out with the planned 4.0 release?
   - Perhaps Java 11 isn't possible for you, in which case we need a 3.x branch for you to commit & merge against
 - Because the change was made against the 3.4.2 commit, but merged against master, this has broken all the CI checks which is unfortunate
   - As above, do we need a 3.x branch?
 - Before merging it would be a good idea to turn the non-JMH benchmark data gathered in to some JMH benchmarks so we can gather more data and get a good idea of the impact
   - I can help here if you like
 - There are concurrency scenarios here which I think should be put down in writing with some JCStress tests, which can help us gauge the frequency of negative capacity values for example
   - I can help here too if you like

bas-k:
In answer to your questions/suggestions:
- We needed this performance improvement for production and the 3.4.2 version was the latest released version. the master is still under development. We run with Java 11, so that is no problem
- We do not need a 3.x branch, I've submitted this pull request so that the code change is maintained in future releases of the official disruptor code base.
- I have no experience with the JMH bechmark tooling. I have read some info about it, but did not have the time yet to play around with it to see if I could benchmark the change with the JMH benchmark.
- Negative capacity could be prevented by returning 0 if the producer overtakes the consumer. This could be considered valid as long as the producer has not published its data, which is guarded by the non-thread-contentious loop for gating in the call to next(n). But I have not looked at all use-cases, because we use only part of the disruptor functionality.

bas-k:
We could copy the SingleProducerSingleConsumer benchmark and update it with the following parameters:

```
@BenchmarkMode(Mode.SampleTime)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
@State(Scope.Benchmark)
...
    @Threads(4)
    @Benchmark
    public void producing()
    {
        for (int cnt=0; cnt < 10_000_000; cnt++)
        {
            long sequence = ringBuffer.next();
            SimpleEvent simpleEvent = ringBuffer.get(sequence);
            simpleEvent.setValue(0);
            ringBuffer.publish(sequence);
        }
    }

```
The `@Thread(4)` annotation as well as the number of messages (10.000.000) depend on the number of cores available to the benchmark. 


Running this code on my local VM with 6 cores gives the following results for the current code and the patched code.

### Current code
Benchmark|Mode|Cnt|Score|Error|Units
--|--|--|--|--|--
MultiProducerSingleConsumer.producing|sample|44|4092.020|± 264.886|ms/op
MultiProducerSingleConsumer.producing:producing·p0.00|sample||3372.220||ms/op
MultiProducerSingleConsumer.producing:producing·p0.50|sample||4083.155||ms/op
MultiProducerSingleConsumer.producing:producing·p0.90|sample||4831.838||ms/op
MultiProducerSingleConsumer.producing:producing·p0.95|sample||5033.165||ms/op
MultiProducerSingleConsumer.producing:producing·p0.99|sample||5737.808||ms/op
MultiProducerSingleConsumer.producing:producing·p0.999|sample||5737.808||ms/op
MultiProducerSingleConsumer.producing:producing·p0.9999|sample||5737.808||ms/op
MultiProducerSingleConsumer.producing:producing·p1.00|sample||5737.808||ms/op
MultiProducerSingleConsumer.producing:·pauses|sample|6|3009.937||ms
MultiProducerSingleConsumer.producing:·pauses.avg|sample||501.656||ms
MultiProducerSingleConsumer.producing:·pauses.count|sample||6.000||#
MultiProducerSingleConsumer.producing:·pauses.p0.00|sample||500.695||ms
MultiProducerSingleConsumer.producing:·pauses.p0.50|sample||501.481||ms
MultiProducerSingleConsumer.producing:·pauses.p0.90|sample||503.316||ms
MultiProducerSingleConsumer.producing:·pauses.p0.95|sample||503.316||ms
MultiProducerSingleConsumer.producing:·pauses.p0.99|sample||503.316||ms
MultiProducerSingleConsumer.producing:·pauses.p0.999|sample||503.316||ms
MultiProducerSingleConsumer.producing:·pauses.p0.9999|sample||503.316||ms
MultiProducerSingleConsumer.producing:·pauses.p1.00|sample||503.316||ms

### Patched code
Benchmark|Mode|Cnt|Score|Error|Units
--|--|--|--|--|--
MultiProducerSingleConsumer.producing|sample|139|1409.301|± 24.601|ms/op
MultiProducerSingleConsumer.producing:producing·p0.00|sample||1270.874||ms/op
MultiProducerSingleConsumer.producing:producing·p0.50|sample||1382.023||ms/op
MultiProducerSingleConsumer.producing:producing·p0.90|sample||1541.407||ms/op
MultiProducerSingleConsumer.producing:producing·p0.95|sample||1579.155||ms/op
MultiProducerSingleConsumer.producing:producing·p0.99|sample||1669.752||ms/op
MultiProducerSingleConsumer.producing:producing·p0.999|sample||1671.430||ms/op
MultiProducerSingleConsumer.producing:producing·p0.9999|sample||1671.430||ms/op
MultiProducerSingleConsumer.producing:producing·p1.00|sample||1671.430||ms/op
MultiProducerSingleConsumer.producing:·pauses|sample||500.171||ms
MultiProducerSingleConsumer.producing:·pauses.avg|sample||500.171||ms
MultiProducerSingleConsumer.producing:·pauses.count|sample||1.000||#
MultiProducerSingleConsumer.producing:·pauses.p0.00|sample||500.171||ms
MultiProducerSingleConsumer.producing:·pauses.p0.50|sample||500.171||ms
MultiProducerSingleConsumer.producing:·pauses.p0.90|sample||500.171||ms
MultiProducerSingleConsumer.producing:·pauses.p0.95|sample||500.171||ms
MultiProducerSingleConsumer.producing:·pauses.p0.99|sample||500.171||ms
MultiProducerSingleConsumer.producing:·pauses.p0.999|sample||500.171||ms
MultiProducerSingleConsumer.producing:·pauses.p0.9999|sample||500.171||ms
MultiProducerSingleConsumer.producing:·pauses.p1.00|sample||500.171||ms

Palmr:
Steps to get this merged:
 - Base it off the current master, your change to `Sequence` is having issues using `UNSAFE` which has been removed for the pending 4.0 release
 - Looks like the results of your JMH benchmark are looking good, commit that too so we can take a look and lend a hand there

Otherwise it's looking like a good change, it's only the fact it was done on top of an old revision and therefore failing the CI checks that's stopping me merging it.

bas-k:
Added a new pull request #356 based on the master branch. This pull request includes

- `UNSAFE` operation removed from  `Sequence` 
- Added the JMH benchmark code for the MultiProducer update.

PR #356, opened 2021-03-15
Increase MultiProducerSequencer performance on getting next sequence (master branch)

bas-k:
The MultiProducerSequencer currently has contention problems when many threads try to get the next sequence. It is not necessary to loop trying to compareAndSet the new sequence number. The producer can allocate sequence numbers upfront as long as it does not return when the consumer has not consumed the slot yet. The up-front allocation is valid because the call to getHighestPublisedSequence is the one that will tell the consumer that a slot has been published by the publisher.

Note: this pull request was first issued as #355 (on the older version). This new pull request is based off the master branch. I've also added the MultiProducerSingleConsumer jmh benchmark test.

### DIFF
--- a/src/jmh/java/com/lmax/disruptor/MultiProducerSingleConsumer.java
+++ b/src/jmh/java/com/lmax/disruptor/MultiProducerSingleConsumer.java
@@ -1,0 +1,74 @@
+package com.lmax.disruptor;
+
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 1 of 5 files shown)


Palmr (approved):

grumpyjames (approved):
This looks a fine change to me, thanks.

We'll take a look at putting a jcstress test or two around this in a separate PR.

PR #357, opened 2021-03-21
Changed expected exception check by JUnit API usage

eas5:
**Problem:**
The idea of this refactoring proposal is to make the exception catching a responsibility of the test framework, which is already provided by its API. Also, without the need of using try/catch blocks, tests can be more straightforward and possibly more comprehensible.

**Solution:**
We propose using JUnit's exception handling to automatically pass/fail the test instead of writing custom exception handling code. In this particular case, JUnit 5 assertThrows() was used to properly handle the expected exception.

**Result:**
_Before:_
```
try
{
    exceptionHandler.handleEventException(causeException, 0L, event);
}
catch (RuntimeException ex)
{
    assertEquals(causeException, ex.getCause());
}
```

_After:_
```
Throwable ex =  assertThrows(RuntimeException.class,() -> {
    exceptionHandler.handleEventException(causeException, 0L, event);
});

assertEquals(causeException, ex.getCause());
```

### DIFF
--- a/src/test/java/com/lmax/disruptor/FatalExceptionHandlerTest.java
+++ b/src/test/java/com/lmax/disruptor/FatalExceptionHandlerTest.java
@@ -31,7 +31,8 @@ public final class FatalExceptionHandlerTest
 
         ExceptionHandler<Object> exceptionHandler = new FatalExceptionHandler();
 
-        Throwable ex =  assertThrows(RuntimeException.class,() -> {
+        Throwable ex =  assertThrows(RuntimeException.class, () ->
+        {
             exceptionHandler.handleEventException(causeException, 0L, event);
         });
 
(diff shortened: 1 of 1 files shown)


Palmr (approved):

Palmr (approved):

Palmr:
Using the JUnit `assertThrows` assertion seems like a good plan.

This PR seems to have only changed one instance of this pattern however? I can see similar try/throw/assert in other tests:
 - `com.lmax.disruptor.TimeoutBlockingWaitStrategyTest#shouldTimeoutWaitFor`
 - `com.lmax.disruptor.FatalExceptionHandlerTest#shouldHandleFatalException` - This is the only one changed so far
 - `com.lmax.disruptor.LiteTimeoutBlockingWaitStrategyTest#shouldTimeoutWaitFor`
 - `com.lmax.disruptor.RingBufferTest#shouldThrowExceptionIfBufferIsFull`
 - `com.lmax.disruptor.SequenceBarrierTest#shouldInterruptDuringBusySpin`
 - `com.lmax.disruptor.SequencerTest#shouldTryNext` 

Before merging this PR it would be good to change all instances to use `assertThrows` where possible, rather than just one.

Palmr:
@eas5 Are you still working on this?

As mentioned in my response before, this PR only replaced 1 out of 6 uses of the same pattern, and this PR will not get merged if it doesn't cover the other states.

If you're not wanting to change the other instances, let me know so I can close this PR and make the complete change myself.

PR #360, opened 2021-04-01
Add method to check if disruptor has been started

ptravers:
I have a use case where I want to check if Disruptor has already been started before calling start

### DIFF
--- a/src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
+++ b/src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
@@ -87,7 +87,8 @@ public class DisruptorTest
     }
 
     @Test
-    public void shouldHaveStartedAfterStartCalled() throws Exception {
+    public void shouldHaveStartedAfterStartCalled() throws Exception
+    {
       if (disruptor.hasStarted())
       {
         fail("Should only be set to started after start is called");
--- a/src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
+++ b/src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
@@ -54,6 +54,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.fail;
 
 @SuppressWarnings(value = {"unchecked"})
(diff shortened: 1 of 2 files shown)


src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
```diff
@@ -86,6 +86,21 @@ public void tearDown() throws Exception
...
+        fail("Should only be set to started after start is called");
+      }
+
+      disruptor.start();
+
+      if (!disruptor.hasStarted())
+      {
+        fail("Should be set to started after start is called");
+      }
+    }
```
Palmr, line 103: This test uses 2-space indentation while all other code in the codebase, and surrounding tests, use 4-space indentation

src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
```diff
@@ -86,6 +86,21 @@ public void tearDown() throws Exception
         executor.joinAllThreads();
     }
 
+    @Test
+    public void shouldHaveStartedAfterStartCalled() throws Exception {
+      if (disruptor.hasStarted())
```
Palmr, line 92: Using `if` and `fail` is an _interesting_ way of writing this test.

Would it not be simpler use `assertFalse(disruptor.hasStarted(), "Should only be set to started after start is called");`?

src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
```diff
@@ -86,6 +86,21 @@ public void tearDown() throws Exception
...
+      {
+        fail("Should only be set to started after start is called");
+      }
+
+      disruptor.start();
+
+      if (!disruptor.hasStarted())
```
Palmr, line 99: Using `if` and `fail` is an _interesting_ way of writing this test.

Would it not be simpler use `assertTrue(disruptor.hasStarted(), "Should be set to started after start is called");`?

Palmr:
Out of curiosity, what's the use case where you need to know if the disruptor has started?
Given `Disruptor::start` is not async if you call start from your code, you know this method will return true.

src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
ptravers: sorry will fix. I'm mostly a scala developer so 2 space has been ingrained into my thumbs!

src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
ptravers: ah yeah! much cleaner! I cribbed from the test below and thought it was notably like golang

ptravers:
sure so checkOnlyStartedOnce, which is called as a precondition in start,  throws an IllegalStateException if start gets called twice. I can't guarantee for various reasons that the code will not call start twice which leaves me with either adding this method or catching the IllegalStateException. I personally feel like the former would make the code cleaner and more descriptive

FWIW I did first try to make it impossible to call start twice but it just wasn't due to some prevailing factors

ptravers:
@Palmr no pressure! but do you have any idea what the timeline will be on this being released?

Palmr:
> @Palmr no pressure! but do you have any idea what the timeline will be on this being released?

@ptravers I don't currently have any set date that I plan on pushing a 4.0 release, and make note that when we do the target version will be JDK11+ too.
If this change is urgent and a snapshot build is no good, let me know and I can think about what we have left on the 4.0 TODO list.

There is another open PR waiting on a second set of review for speeding up MultiProducerSequencers that I would also like to get out sooner rather than later, so a potential for a 3.4.4 release with those change in too at some point.

ptravers:
@Palmr I am fine to wait til the release of 4.0 or 3.4.4 and we are compitable with JDK 11+ fortunately. Thanks so much for response!

PR #362, opened 2021-04-05
Use platform logging API in FatalExceptionHandler and IgnoreExceptionHandler

A248:
Changing from java.util.logging to the [Platform Logging API](https://openjdk.java.net/jeps/264) is preferable now that this library is targeting JDK 11.

* Removes the dependency on java.util.logging, which some modular applications may want to avoid.
* Both java.util.logging and System.Logger can be redirected to a logging framework of choice. However, doing so is far more clean, straightforward, and effective for System.Logger. (I would wager performant too, given jul's creation of LogRecord objects)
* By default, the System.Logger API routes to java.util.logging, so for the majority of users this will be a non-breaking change and the behavior will be exactly the same. It is listed as a breaking change because:
  * FatalExceptionHandler's constructor taking a Logger has been removed.
  * It affects the highly unlikely setup where the user has implemented the system logging API to use a certain logging framework but has not done the same for java.util.logging.
* This may coincidentally solve some initialization cycles in Log4j between log4j-core and log4j-jul, see https://github.com/apache/logging-log4j2/pull/447 . That is not the main benefit of this change however.

### DIFF
--- a/src/main/java/com/lmax/disruptor/IgnoreExceptionHandler.java
+++ b/src/main/java/com/lmax/disruptor/IgnoreExceptionHandler.java
@@ -15,43 +15,32 @@
  */
 package com.lmax.disruptor;
 
-import java.util.logging.Level;
-import java.util.logging.Logger;
+import java.lang.System.Logger.Level;
+import java.lang.System.Logger;
 
 /**
- * Convenience implementation of an exception handler that using standard JDK logging to log
- * the exception as {@link Level}.INFO
+ * Convenience implementation of an exception handler that uses the standard JDK logging
+ * of {@link System.Logger} to log the exception as {@link Level}.INFO
  */
 public final class IgnoreExceptionHandler implements ExceptionHandler<Object>
 {
-    private static final Logger LOGGER = Logger.getLogger(IgnoreExceptionHandler.class.getName());
-    private final Logger logger;
-
-    public IgnoreExceptionHandler()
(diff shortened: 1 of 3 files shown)


src/main/java/com/lmax/disruptor/IgnoreExceptionHandler.java
```diff
@@ -15,43 +15,32 @@
...
-    {
-        this.logger = logger;
-    }
+    private static final Logger LOGGER = System.getLogger(IgnoreExceptionHandler.class.getName());
 
     @Override
     public void handleEventException(final Throwable ex, final long sequence, final Object event)
     {
-        logger.log(Level.INFO, "Exception processing: " + sequence + " " + event, ex);
+        LOGGER.log(Level.INFO, "Exception processing: " + sequence + " " + event, ex);
```
Palmr, line 32: Given the unknown complexity of `event::toString` I wonder if, on the `IgnoreExceptionHandler` at least, we should use the supplier variant of `LOGGER::log`?
Then if using the ignore handler, and you have a log level higher above `INFO` it won't spend time building that string?

A248: Good idea. I can't think of any serious downsides to using the supplier, so I'll be consistent and use that in both handlers.

Palmr (approved):

PR #364, opened 2021-04-09
Add rewind batch feature to the BatchEventProcessor.

isolgpus:
When processing an event, if an exception is thrown that could be recoverable via outside influence. Throw a RewindableException that will replay all the events in the current batch.


src/main/java/com/lmax/disruptor/NanosecondRewindPauseStrategy.java
```diff
@@ -0,0 +1,20 @@
+package com.lmax.disruptor;
+
+import java.util.concurrent.locks.LockSupport;
+
+public class NanosecondRewindPauseStrategy implements RewindPauseStrategy
+{
+
+    private final long nanoSecondPauseTime;
+
+    public NanosecondRewindPauseStrategy(long nanoSecondPauseTime)
```
Palmr, line 10: Checkstyle has been updated and requires final on all parameters

src/main/java/com/lmax/disruptor/RewindableException.java
```diff
@@ -0,0 +1,9 @@
+package com.lmax.disruptor;
+
+public class RewindableException extends RuntimeException
+{
+    public RewindableException(Throwable cause)
```
Palmr, line 5: Checkstyle has been updated and requires final on all parameters

src/main/java/com/lmax/disruptor/BatchEventProcessor.java
```diff
@@ -180,6 +198,11 @@ private void processEvents()
                     break;
                 }
             }
+            catch (final RewindableException e)
```
Palmr, line 201: In the event of a permanent exception, which may be classed as rewindable is there any idea how you're meant to track how many attempts it has at processing the batch?

Are you meant to encapsulate that in `RewindPauseStrategy::pause`?
I wonder if the `RewindableException e` should be passed to the pause strategy, and if it should be geared a little more towards more than just pausing, e.g. logging that exception?

Palmr (commented):
Checkstyle for final, and slight question on what to do with the RewindableException handler.

Otherwise seems a fairly small and potentially useful change with nice tests.

src/main/java/com/lmax/disruptor/BatchEventProcessor.java
isolgpus: I have generified the RewindPauseStrategy to a RewindStrategy. The user is free to provide a custom strategy and decide whether the batch should rewind or delegate the exception to the exception handler. This will give it more usefulness than just pausing.

isolgpus:
Documentation has been added to give an explanation on a useful use case for batch rewind.

Issue #365, opened 2021-04-12
Missing Javadoc


Palmr:
I was looking at tightening the Checkstyle rules but had to comment out rules checking for Javadoc on publicly scoped types, methods and variables.

The following is a list of currently missing Javadoc which we should get cleaned up for a 4.0 release:

```
[ant:checkstyle] [ERROR] /home/nick/git-projects/disruptor/src/main/java/com/lmax/disruptor/BatchStartAware.java:8:5: Missing a Javadoc comment. [JavadocMethod]
[ant:checkstyle] [ERROR] /home/nick/git-projects/disruptor/src/main/java/com/lmax/disruptor/DataProvider.java:18: Missing a Javadoc comment. [JavadocType]
[ant:checkstyle] [ERROR] /home/nick/git-projects/disruptor/src/main/java/com/lmax/disruptor/DataProvider.java:20:5: Missing a Javadoc comment. [JavadocMethod]
[ant:checkstyle] [ERROR] /home/nick/git-projects/disruptor/src/main/java/com/lmax/disruptor/EventPoller.java:15: Missing a Javadoc comment. [JavadocType]
[ant:checkstyle] [ERROR] /home/nick/git-projects/disruptor/src/main/java/com/lmax/disruptor/EventPoller.java:17:9: Missing a Javadoc comment. [JavadocMethod]
[ant:checkstyle] [ERROR] /home/nick/git-projects/disruptor/src/main/java/com/lmax/disruptor/EventPoller.java:20: Missing a Javadoc comment. [JavadocType]
[ant:checkstyle] [ERROR] /home/nick/git-projects/disruptor/src/main/java/com/lmax/disruptor/EventPoller.java:22:9: Missing a Javadoc comment. [JavadocVariable]
[ant:checkstyle] [ERROR] /home/nick/git-projects/disruptor/src/main/java/com/lmax/disruptor/EventPoller.java:22:21: Missing a Javadoc comment. [JavadocVariable]
[ant:checkstyle] [ERROR] /home/nick/git-projects/disruptor/src/main/java/com/lmax/disruptor/EventPoller.java:22:29: Missing a Javadoc comment. [JavadocVariable]
[ant:checkstyle] [ERROR] /home/nick/git-projects/disruptor/src/main/java/com/lmax/disruptor/EventPoller.java:25:5: Missing a Javadoc comment. [JavadocMethod]
[...]

PR #366, opened 2021-04-12
Supply some missing javadoc commentary

grumpyjames:
Work towards https://github.com/LMAX-Exchange/disruptor/issues/365


src/main/java/com/lmax/disruptor/EventProcessor.java
```diff
@@ -38,5 +38,9 @@
      */
     void halt();
 
+    /**
+     * @return whether this event processor is running or not
+     * Implementations should ideally return true only when the associated thread is idle.
```
Palmr, line 43: > Implementations should ideally return true only when the associated thread is idle

I thought `isRunning` returns true only when the associated thread is _NOT_ idle?

grumpyjames: booleans will be booleans. Good catch, thank you.

src/main/java/com/lmax/disruptor/Sequenced.java
```diff
@@ -1,5 +1,9 @@
 package com.lmax.disruptor;
 
+/**
+ * Operations related to the sequencing of items in a {@link RingBuffer}.
+ * See the two child interfaces, {@link Sequenced} and {@link EventSequencer} for more details.
```
Palmr, line 5: Child interface would be Sequence**r**

Palmr (changes requested):
Looks good, though think that link to `Sequenced` from `Sequenced` is incorrect.

Palmr (approved):

PR #371, opened 2021-04-13
Fix #370: Logger instances are lazily initialized

### LINKED ISSUE #370: Provide a way to register a custom ExceptionHandler without initializing java.util.logging
We use the disruptor library heavily in log4j2 for [fully asynchronous logging](https://logging.apache.org/log4j/2.x/manual/async.html). Unfortunately, there are some edge cases which may result in a deadlock on initialization. See [LOG4J2-2965](https://issues.apache.org/jira/browse/LOG4J2-2965) with a reproducer here: https://github.com/apache/logging-log4j2/pull/447

Ideally we would be able to provide a custom exception handler that avoids initializing the FatalExceptionHandler (thus JUL) unnecessarily: https://github.com/LMAX-Exchange/disruptor/blob/2d4762336e2ad6d475dcf0310a980c50fd5bff0b/src/main/java/com/lmax/disruptor/FatalExceptionHandler.java#L26-L28

This could be done by lazily creating the FatalExceptionHandler if no other handler has been provided when the ring buffer is started, although there are subtler approaches to reduce code change, for example wrapping the `FatalExceptionHandler` Logger instance within a static inner class to delay initialization until the logger field is first accessed based on [jls-12.4.2](https://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.4.2).

Would you accept a contribution as described above? Thanks!

carterkozak:
Default java loggers are no longer initialized when custom
ExceptionHandler instances are provided. This allows logging
frameworks to delegate to disruptor without risking deadlocks.

Implementation notes:

This introduces a new public utility class `ExceptionHandlers` which exposes a singleton `ExceptionHandler<Object>` that can be lazily initialized. This allows the handler to be centrally controlled and shared between `BatchEventProcessor` and `ExceptionHandlerWrapper`.

The `ExceptionHandler` instances are stored in non-volatile fields, so I opted to null check and produce the default singleton instead of attempting to initialize the field to some default just prior to initialization. This way even unexpected concurrent initial accesses should not be able to produce different behavior if this patch is applied.

`BatchEventProcessor` is performance sensitive, so I opted to structure changes in such a way that method size will not grow, in case it's optimized by size for inlining. Exception handling generally occurs out of hot paths, so additional method call overhead should be negligible, and the calling method size should be slightly lower, avoiding the `exceptionHandler` object lookup bytecode. As an aside, the `BatchEventProcessor` hot loop may benefit from a single final field read outside of the loop to avoid unnecessary indirection because `-XX:+TrustFinalNonStaticFields` still isn't the default JVM behavior.

### DIFF
--- a/src/jcstress/java/com/lmax/disruptor/LoggerInitializationStress.java
+++ b/src/jcstress/java/com/lmax/disruptor/LoggerInitializationStress.java
@@ -1,0 +1,150 @@
+package com.lmax.disruptor;
+
+import com.lmax.disruptor.dsl.Disruptor;
+import com.lmax.disruptor.dsl.ProducerType;
+import com.lmax.disruptor.util.Util;
+import org.openjdk.jcstress.annotations.Actor;
+import org.openjdk.jcstress.annotations.JCStressTest;
+import org.openjdk.jcstress.annotations.Mode;
+import org.openjdk.jcstress.annotations.Outcome;
+import org.openjdk.jcstress.annotations.Signal;
+import sun.misc.Unsafe;
+
+import java.lang.reflect.Field;
+import java.util.concurrent.Executors;
+import java.util.logging.LogManager;
+
+import static org.openjdk.jcstress.annotations.Expect.ACCEPTABLE;
+import static org.openjdk.jcstress.annotations.Expect.FORBIDDEN;
+
+/**
+ * Validate that creating a {@link Disruptor} instance with a custom {@link ExceptionHandler} does not
+ * initialize the logging framework. Using JCStress is a stretch, we're not validating a race condition,
+ * rather that the logging framework is not started. This type of test doesn't work well in unit tests
+ * because it requires JVM isolation, and any interactions with a logger invalidate the test.
+ */
+@JCStressTest(Mode.Termination)
+@Outcome(id = "TERMINATED", expect = ACCEPTABLE, desc = "Logger has not been initialized")
+@Outcome(
+        id = {"STALE", "ERROR"},
+        expect = FORBIDDEN,
+        desc = "Logger has been initialized. This has the potential to cause " +
+                "deadlocks when disruptor is used within logging frameworks")
+public class LoggerInitializationStress
+{
+    private static volatile boolean logManagerInitialized;
+    private static volatile boolean disruptorInitialized;
+    static
+    {
+        System.setProperty("java.util.logging.manager", DisruptorLogManager.class.getCanonicalName());
+    }
+
+    @Actor
+    public void actor() throws Exception
+    {
+        // Wait for a the Disruptor instance to be initialized
+        while (!disruptorInitialized)
+        {
+        }
+        // Validate state
+        if (logManagerInitialized)
+        {
+            throw new RuntimeException("Expected the LogManager to be uninitialized");
+        }
+        // Use Unsafe to determine if the LogManager has been initialized.
+        // Accessing the LogManager normally using LogManager.getLogManager
+        // results in initialization which modifies static state and prevents
+        // subsequent runs from executing successfully.
+        Field managerField = LogManager.class.getDeclaredField("manager");
+        Unsafe unsafe = Util.getUnsafe();
+        Object managerBase = unsafe.staticFieldBase(managerField);
+        long managerOffset = unsafe.staticFieldOffset(managerField);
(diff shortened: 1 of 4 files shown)


carterkozak:
This could be implemented with fewer lines of code, but the code would be less obvious and incur slightly higher startup cost (e.g. unnecessary object creation). I'd be happy to implement this in another way if you prefer. Thanks!

carterkozak:
This is the simpler alternative for comparison: https://github.com/carterkozak/disruptor/commit/e350a7fe309eaa149330d965eb86740b8e043eb0

Palmr:
Thanks for the PR @carterkozak all looks good.

It would be nice to have a test that fails if the loading of the logger in `FatalExceptionHandler` or elsewhere changed in future to be non-lazy.
But I'm not sure what such a test would look like and if it would be possible/sensible to do?

Did you have any thoughts on this?
I saw in the earlier issue there's a JMH benchmark which was used to try and catch a deadlock on the log4j side, I wondered if this could be of use? Perhaps should be a JCStress test?
Setup just in Disruptor might be too heavy though, as I assume it would need some kind of small test logger framework too?

carterkozak:
Thanks, @Palmr. I wanted to implement a simple unit test for this, but it's difficult to do in a reasonable way.

The JMH reproducer requires a logging framework which is both initialized via the system logger as well as an implementation which relies on disruptor.

The closest idea I had was to run a test with a custom delegating LogManager implementation (initialized on the first use) which tracks interactions, that way we can validate whether or not the system logger has been accessed. This approach relies on static state (loggers themselves tend to be static state) so the unit test would be required to run in its own fork, otherwise would potentially produce incorrect data. Also, junit 5 initializes a JUL logger in setup, which would invalidate the test. A unit test could reuse the current classpath and launch another JVM via a test `main` method, but this style test is fragile.
I can put some time into experimenting with such a test if you like, but it's likely to be difficult to maintain over time. What do you think?

Palmr:
I would advise doing this kind of test with JUnit or JMH involved for the reasons you've mentioned.

A JCStress test in Termination mode (e.g. https://github.com/openjdk/jcstress/blob/010fc224643490dd13519ede33eda9ff0c276fc3/jcstress-samples/src/main/java/org/openjdk/jcstress/samples/APISample_03_Termination.java) would potentially be a better test framework for writing this kind of test where we're looking for a fairly rare occurrence of a deadlock between threads.

Though as I mentioned before it may be a bit too heavyweight of a test to have in the Disruptor code, with the need for a custom delegating LogManager implementation and potential other setup required.

The patch seems fair, mostly unnoticeable in normal circumstances and in theory it should fix the problems initially mentioned in the issue.
A test would be nice to give me more tangible proof that it fixes the original issue, but I can run this patched disruptor through the original JMH-based reproducer to get my fix of seeing a test flip from red to green. (I take it you've already done this?)

I'm thinking I should merge this PR in its current state and leave any potential tests as a "nice to have" that could come in a later PR if time is found and results of experimentation with testing frameworks yields results.

carterkozak:
I've put together a working JCStress test here: #374
It's not all that bad, but it's a stretch to use JCStress in this case -- it's a convenient place to run validation on a forked JVM.

In the JCStress case I'm not attempting to reproduce the deadlock, but validate that the root cause of the deadlock (logger initialization when a Disruptor instance is created and started) no longer occurs. That PR fails on master, but passes when this commit is present.
Would you like me to add that commit to this PR?

Palmr:
> Would you like me to add that commit to this PR?

I think so.

I like the idea that there is _something_ that shows this commit does indeed fix the root cause of that deadlock.
If you add the test to this PR I'll merge it all, and can then close #374

carterkozak:
Sounds good, thanks! I've added that commit to this PR and added inline comments describing the `Unsafe` piece of the test which validates logging has not been initialized.

Palmr:
Seems the CI checks hadn't run until I allowed it to, not something I've seen GitHub do before.

But now it has run it seems you've fallen foul of recently-added checkstyle rules: missing a `@return` in your JavaDoc at `ExceptionHandlers.java:23`

Once that's added I can merge :)

carterkozak:
I see, after a rebase I was able to reproduce the checkstyle failure. Updated.

carterkozak:
Thanks for the review :-)

PR #373, opened 2021-04-21
GHA: Perform "quick" jcstress run instead of "sanity"

shipilev:
I noticed that GHA workflow runs jcstress "sanity" mode as the commit-level check. "sanity" mode is really for jcstress infrastructure testing, as it spends almost no time in the tests themselves. If we want to run the fast testing that spend some time in test code, we are better off doing "quick" runs.

Downside: it takes about 7 minutes to run. This can be made better to culling the number of JVM configuration the tests run.

### DIFF
--- a/.github/workflows/jcstress-quick.yml
+++ b/.github/workflows/jcstress-quick.yml
@@ -1,4 +1,4 @@
-name: JCStress Testing - commit-level sanity check
+name: JCStress Testing - commit-level quick check
 
 on: [ push, pull_request ]
 
@@ -18,9 +18,9 @@ jobs:
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
       - name: Build with Gradle
-        run: ./gradlew jcstress -Pmode=sanity
+        run: ./gradlew jcstress -Pmode=quick
       - name: Archive artifacts
         uses: actions/upload-artifact@v2
         with:
-          name: jcstress-report-sanity
+          name: jcstress-report-quick
           path: build/reports/jcstress


Palmr (approved):

Palmr:
> Downside: it takes about 7 minutes to run. This can be made better to culling the number of JVM configuration the tests run.

This is probably something we should do, 7 minute checks on commit is too long.
The commit check was mostly to check that the tests _can_ run, more than it was there to check they might catch any concurrency issues.
I think ~1-2 minutes would probably be my limit for runtime of a commit jcstress check.

There is another action running every other week which runs in tough mode to try and catch issues, though that has started hitting a 6hr timeout in GitHub unfortunately.
I should probably look at running these on self-hosted GHA runners which will let us have more time to run tests and the ability to run on ARM too.

shipilev:
> The commit check was mostly to check that the tests _can_ run, more than it was there to check they might catch any concurrency issues.

Oh, then "sanity" is exactly what you want here. You can close this PR without integration :)

jcstress 0.9 would make runs longer, as it comes with split compilation and actor affinity, that increases the configurations per test quite significantly. So, "tough" would become even longer. Additionally, "sanity" would become a tad longer as well, as it would fork at least once per test, in contrast with current harness only running it in the same JVM. 

So, "sanity" for commit-level check is fine. Weekly runs are probably good enough with "default", or maybe with fork count bumped up just a little.

Self-hosted runners would be better, mostly because jcstress 0.9 would try to run the tests across/within threads, cores, packages, so the more interesting the host is, the more new behaviors would be found. GHA runners are two-cpu instances, which is not that interesting.

Palmr:
Perhaps `sanity` is the way to go for GHA push triggers, but `quick` could be the thing to enable for pull_request triggers, where a little more checking is worth it?

I shall have to check out 0.9 and experiment once I have a test runner with more than a few cores available.

This is the first time I've used JCStress so all comments, advice, and information welcome.
I suppose I should join the mailing lists mentioned in https://github.com/openjdk/jcstress/#interpreting-the-results

shipilev:
> Perhaps `sanity` is the way to go for GHA push triggers, but `quick` could be the thing to enable for pull_request triggers, where a little more checking is worth it?

I think the weekly schedule with `default` is good enough, assuming it runs before release :)

> This is the first time I've used JCStress so all comments, advice, and information welcome.

I looked at current tests briefly, and they are fine.

Palmr:
After more thinking time, I'm coming around to it and now thinking I should merge this PR.

I'm getting set up with provisioning GHA self-hosted runners which should run quick mode tests in a couple of minutes, so the concern over a 7-minute wait time on CI for each commit will be avoided.

Better provisioned, faster, CI runners should make jcstress 0.9 easier to handle too.
I could potentially change the current `tough` run every other week to a weekly `default` run too.

At some point I will make sure a release is an automated CI process, rather than the current manual method. Then I shall enforce a clean jcstress run first! For now it remains a hopeful assumption that it passes CI :)

PR #374, opened 2021-04-21
JCStress test to validate that loggers aren't initialized

carterkozak:
Validate that creating a Disruptor instance with a custom ExceptionHandler does not
initialize the logging framework. Using JCStress is a stretch, we're not validating a race condition,
rather that the logging framework is not started. This type of test doesn't work well in unit tests
because it requires JVM isolation, and any interactions with a logger invalidate the test.

When disruptor is used within a logging framework, it's crucial that loggers are not accessed internally on initialization because such access is likely to result in deadlocking.

See:
[LOG4J2-2965](https://issues.apache.org/jira/browse/LOG4J2-2965), #370, #371

This JCStress test currently fails on master, but passes when #371 is applied.


carterkozak:
I've added this commit to #371

PR #375, opened 2021-04-23
General improvements to the documentation

swarren12:
A PR for general improvements to the documentation.

So far:

- splitting out the examples so that we can prove they compile (spoilers: some previously didn't)
- highlighting important pieces of information using admonitions
- general rework of the "Using the Disruptor" section

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ plugins {
     id 'checkstyle'
     id 'idea'
     id "io.github.reyerizo.gradle.jcstress" version "0.8.11"
-    id 'org.asciidoctor.jvm.convert' version '3.1.0'
+    id 'org.asciidoctor.jvm.convert' version '3.3.0'
     id "me.champeau.jmh" version "0.6.5"
 }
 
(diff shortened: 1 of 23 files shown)


swarren12:
One thing I'm not too happy about right now is that Gradle is enforcing checkstyle on the examples, which is nice in theory but it causes some lines to wrap in inconvenient ways... not sure if we should reformat or just turn it off?

Palmr:
> it causes some lines to wrap in inconvenient ways

Is there a specific example of this?


You can add a suppression to `config/checkstyle/suppress.xml` e.g.

```xml
<suppress checks="LineLength" files="examples[\\/].*"/>
```

swarren12:
>  > it causes some lines to wrap in inconvenient ways

> Is there a specific example of this?

It's mostly in the method signatures of things where checkstyle seems to want parameters to be final, e.g.:

```java
public static void handleEvent(final LongEvent event, final long sequence, final boolean endOfBatch)
```

Because of the formatting, this ends up rendered in the HTML as:

```java
        public static void handleEvent(final LongEvent event, final long sequence, final boolean
endOfBatch)
```

Palmr:
Feels like what you want is checkstyle to have a smaller `LineLength` rule for code under `src/examples` to play nicer with the `max-width: 62.5em` on `#content` in the asciidoc theme we have?

PR #383, opened 2021-06-11
Catch exception when delegate exception handler process.

horizonzy:
If someone define a exception handler which will throw exception when process, catch it to avoid program exit.

### DIFF
--- a/src/main/java/com/lmax/disruptor/dsl/ExceptionHandlerWrapper.java
+++ b/src/main/java/com/lmax/disruptor/dsl/ExceptionHandlerWrapper.java
@@ -43,7 +43,7 @@ public class ExceptionHandlerWrapper<T> implements ExceptionHandler<T>
     {
         try
         {
-            getExceptionHandler() .handleOnShutdownException(ex);
+            getExceptionHandler().handleOnShutdownException(ex);
         }
         catch (Exception e)
         {
--- a/src/main/java/com/lmax/disruptor/dsl/ExceptionHandlerWrapper.java
+++ b/src/main/java/com/lmax/disruptor/dsl/ExceptionHandlerWrapper.java
@@ -15,19 +15,40 @@ public class ExceptionHandlerWrapper<T> implements ExceptionHandler<T>
     @Override
     public void handleEventException(final Throwable ex, final long sequence, final T event)
     {
-        getExceptionHandler().handleEventException(ex, sequence, event);
+        try
+        {
+            getExceptionHandler().handleEventException(ex, sequence, event);
+        }
+        catch (Exception e)
+        {
+            e.printStackTrace();
+        }
     }
 
     @Override
     public void handleOnStartException(final Throwable ex)
     {
-        getExceptionHandler().handleOnStartException(ex);
+        try
+        {
+            getExceptionHandler().handleOnStartException(ex);
+        }
+        catch (Exception e)
+        {
+            e.printStackTrace();
+        }
     }
 
     @Override
     public void handleOnShutdownException(final Throwable ex)
     {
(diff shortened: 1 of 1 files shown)


Palmr:
Hi @horizonzy , thank you for taking the time to send us a pull request.

Unfortunately I do not think this is a change we would want to make.

The default exception handler is the `FatalExceptionHandler` and it's main job is to be fatal, and throw an exception after logging.
Wrapping the code that calls this with a try-catch, handling that exception, would go against that purpose.

If you wanted an exception handler to not throw an exception, to avoid the program exiting, then you can define an exception handler that does such a thing, like the `IgnoreExceptionHandler` for example.

Exception handlers throwing exceptions is an exceptional event, and should be treated accordingly.

---

Also, from a code-review point of view, I see you used `catch (Exception e)` and `e.printStackTrace()`, both of which I would probably advise against.
In your case it seems you want to try and catch any possibly problem from the handler, and log the error. Catching `Throwable` might have been preferred given that covers `Exception` as well as `Error` which could have been thrown.

`e.printStackTrace()` is generally bad practice as it's going to print that stack trace to `System.err` and where that gets redirected is out of the libraries control.
If we want to log the stack trace we should use the logging framework, and you can see that's what we do in the `FatalExceptionHandler` and `IgnoreExceptionHandler`.

horizonzy:
> Hi @horizonzy , thank you for taking the time to send us a pull request.
> 
> Unfortunately I do not think this is a change we would want to make.
> 
> The default exception handler is the `FatalExceptionHandler` and it's main job is to be fatal, and throw an exception after logging.
> Wrapping the code that calls this with a try-catch, handling that exception, would go against that purpose.
> 
> If you wanted an exception handler to not throw an exception, to avoid the program exiting, then you can define an exception handler that does such a thing, like the `IgnoreExceptionHandler` for example.
> 
> Exception handlers throwing exceptions is an exceptional event, and should be treated accordingly.
> 
> Also, from a code-review point of view, I see you used `catch (Exception e)` and `e.printStackTrace()`, both of which I would probably advise against.
> In your case it seems you want to try and catch any possibly problem from the handler, and log the error. Catching `Throwable` might have been preferred given that covers `Exception` as well as `Error` which could have been thrown.
> 
> `e.printStackTrace()` is generally bad practice as it's going to print that stack trace to `System.err` and where that gets redirected is out of the libraries control.
> If we want to log the stack trace we should use the logging framework, and you can see that's what we do in the `FatalExceptionHandler` and `IgnoreExceptionHandler`.



> Hi @horizonzy , thank you for taking the time to send us a pull request.
> 
> Unfortunately I do not think this is a change we would want to make.
> 
> The default exception handler is the `FatalExceptionHandler` and it's main job is to be fatal, and throw an exception after logging.
> Wrapping the code that calls this with a try-catch, handling that exception, would go against that purpose.
> 
> If you wanted an exception handler to not throw an exception, to avoid the program exiting, then you can define an exception handler that does such a thing, like the `IgnoreExceptionHandler` for example.
> 
> Exception handlers throwing exceptions is an exceptional event, and should be treated accordingly.
> 
> Also, from a code-review point of view, I see you used `catch (Exception e)` and `e.printStackTrace()`, both of which I would probably advise against.
> In your case it seems you want to try and catch any possibly problem from the handler, and log the error. Catching `Throwable` might have been preferred given that covers `Exception` as well as `Error` which could have been thrown.
> 
> `e.printStackTrace()` is generally bad practice as it's going to print that stack trace to `System.err` and where that gets redirected is out of the libraries control.
> If we want to log the stack trace we should use the logging framework, and you can see that's what we do in the `FatalExceptionHandler` and `IgnoreExceptionHandler`.

Thanks for your suggestion.

PR #386, opened 2021-06-21
Fix siteUrl

TobiX:
The current URL (at github.com) only shows a 404 error.

While I was at it I switched the Apache Licence URL to HTTPS


Palmr:
Hi @TobiX the change for the `siteUrl` seems good and useful.

I'd be tempted to revert the change to Apache license however as you've only changed 1 instance out of 138.
And that feels like a separate task for checking/updating the license.

TobiX:
Good point. Fixed this pull request to only contain the siteUrl change

TobiX:
Created #387 for switching all URLs to HTTPS...

Palmr (approved):

Issue #398, opened 2021-10-13
Improve `BatchEventProcessor` current method of detecting EventHandler extensions via `instanceof`


Palmr:
There's a number of `instanceof`'s inside `BatchEventProcessor` which could probably be avoided if the `EventHandler` interface instead had a few more methods on it?

They could have default implementations to reduce the migration effort.

PR #399, opened 2021-10-24
Reduce the amount of `instanceof`

Palmr:
As mentioned in #398 the `BatchEventProcessor` uses `instanceof` (along with casting and null checks) a few times to detect `eventHandler` which may implement some "extension" interfaces.

This use of `instanceof` could be removed by making the interface `EventHandler` extend the extension interfaces, so all implementations of `EventHandler` will be known to have those extension methods on them.

Since Disruptor 4.x is targeting Java 11+ we can now use default methods in those interfaces to reduce the amount of code change required.

TheEvent Handler "extension" interfaces are:

 - `BatchStartAware`
 - `TimeoutHandler`
 - `SequenceReportingEventHandler`
 - `LifecycleAware`

This also reduces the need for `instanceof` and casting from `AggregateEventHandler` as well as `BatchEventProcessor`

I have run the old performance tests via the PerfTest jar and all results were equivalent or faster with this commit.
There are no dedicated JMH tests.


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
     - Minimum Java version now 11
     - Issue #323 - `WorkerPool` and `WorkProcessor` have been removed, no more `Disruptor::handleEventsWithWorkerPool`
     - `Disruptor` constructors using Executor have been removed. Use ThreadFactory instead.
+    - Rolled up event handling extnsion interfaces on to EventHandler:
```
grumpyjames, line 9: typo: extnsion

grumpyjames:
I think my only concern here is, *sigh of the ancients*, performance of that polymorphic dispatch, specifically for BatchStartAware, because we call that one in a mildly heated fashion.

I'm assuming that in our benchmarks, we've got _exactly one_ EventHandler instance, and so we'll likely monomorph (and maybe inline) the no-op defaults. It _might_ be worth trying a benchmark out where multiple EH implementations are available to see if its any slower?

Palmr:
I created a pair of benchmarks to test one vs. multiple implementations of `EventHandler`, and one handler that just overrode `onEvent` and another that overrode all the methods.
No difference was measured with a benchmark that ran 1m events through and waited for the consumer to catch up.

PR #408, opened 2021-12-29
Refactor pacakge structure of main module.

RocMarshal:
Refactor pacakge structure.


sea36:
What is the reason for proposing this?

Rearranging the package structure is a breaking change to the API, so needs a strong justification.

RocMarshal:
@sea36 Glad to get your attention.
In my limited read,

- The location of the class file inside the package is more clear
- Easier to read the code
- Clear class arrange layout woud be good for iterating and evolving.

but,
- Will destroy the current api structure and visibility
- If there is a lack of adequate test cases, potential errors will accompany

Thank you.

Palmr:
Thank you for the proposal, and the explanation.

We are at a point of release v4 which is a major version where the API is allowed to have breaking changes. So it's not out of the question we _could_ merge this.

I like the change in theory but I am on the fence about if the refactor every project that uses the Disruptor will have to do is worth it.

Happy to gather more feedback from others on this, if anyone (@sea36 @grumpyjames @swarren12) wants to comment or leave a :+1: or :-1: feel free.

grumpyjames:
Here's the question I'd ask: does this change make it easier or harder to reason about what this library's public interface is, to an external user?

I think the answer is harder, because to create this package structure we've ended up making _more_ classes public.

As such I'm a no.

sea36:
My feeling is that the amount of change / impact on users isn't worth the benefits.

If it was just lesser used parts of the library I might feel differently, but this change moves some really core parts around, so pretty much everyone is going to be forced to make changes when upgrading.

Overall, I'm a no.

Palmr:
That's two :-1:  and I have no strong urge to disagree, so I will close this PR & close #409 too

Issue #405, opened 2021-12-14
Something optimize suggestion.


SteafanMrZhou:
**Describe the bug**
When i studying source code of Disruptor,i find some optimize bugs,so i wana solve it.
eg.No @ Override、No Complete annotation.
Can i fix it? Cause i want be a contributor for Disruptor,thanks.

Palmr:
Yes, please raise a pull request and we will review.

Palmr:
Closing this issue while waiting for a PR to show up

SteafanMrZhou:
@Palmr Hi,I'm so sorry to delay raising a pull request for issue#405,cause i'm very busy a period ago.
Awaiting your review,thanks.

PR #410, opened 2022-01-05
A PR for issue#405.

SteafanMrZhou:
I'm so sorry to delay raising a pull request for issue#405,cause i'm very busy a period ago.
I annotated some method overloads by @Override,it is necesseary for java project.
There are others like this：curly braces should follow the class name immediately, not on another line,etc.
Awaiting your review,thanks.

### DIFF
--- a/src/examples/java/com/lmax/disruptor/examples/longevent/LongEventFactory.java
+++ b/src/examples/java/com/lmax/disruptor/examples/longevent/LongEventFactory.java
@@ -5,6 +5,7 @@ import com.lmax.disruptor.EventFactory;
 // tag::example[]
 public class LongEventFactory implements EventFactory<LongEvent>
 {
+    @Override
     public LongEvent newInstance()
     {
         return new LongEvent();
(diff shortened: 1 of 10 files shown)


SteafanMrZhou:
#405

Palmr (approved):
These all seem fine to me

Palmr:
All the places you have added the `@Override` annotation seem appropriate and I'm fairly surprised they were missing in the first place. Good spot :+1: 

---

A couple of points on your comments though:

> it is necesseary for java project

The `@Override` annotation is good practice as it makes it easy to spot methods that are overriding and the compiler can throw an error if you annotate a method with it that is not actually overriding a method on a parent class/interface.

But it is not really necessary, if you do not have it, the bytecode is no different and the running of your application is no different. So `@Override` annotations being being missing is not a bug and adding them is not an optimisation.

But I would agree it's better to have them than not have them.

> There are others like this：curly braces should follow the class name immediately, not on another line,etc.

This change would be a style change only, and would go against the checkstyle rules currently in this project.

Do not spend time on this as there is no benefit and the PR will be rejected.

PR #414, opened 2022-03-02
Assert that only a single thread accesses a SingleProducerSequencer.

nicholassm:
If a developer uses the wrong ProducerType, it's hard to detect.

This commit adds a check using assertions so a developer can fail in
test instead of being burnt in Production.
(There will be no runtime cost in Production.)

As can be seen from the commit, there was actually a test case that
started failing because a SingleProducerSequencer was accessed by two
threads...

### DIFF
--- a/src/main/java/com/lmax/disruptor/SingleProducerSequencer.java
+++ b/src/main/java/com/lmax/disruptor/SingleProducerSequencer.java
@@ -18,6 +18,8 @@ package com.lmax.disruptor;
 import com.lmax.disruptor.util.Util;
 
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.locks.LockSupport;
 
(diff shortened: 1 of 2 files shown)


nicholassm:
Hi, I realized that we actually don't even have to pay the cost of an extra field (if assertions are not enabled) by using a private static inner class that will only be loaded if assertions are enabled. It also encapsulates the assertion logic in a better way - rather than polluting the business logic.

Palmr:
I'm planning to take a deeper look at this tomorrow, but on the surface it seems like a good idea and implementation.

> Btw: Let me know if you want me to squash the commits into a single commit.

Don't worry about it

nicholassm:
Hi,

Is there anything you want me to do? Or that is missing?

Kind regards,
    Nicholas

Palmr:
Sorry @nicholassm I started looking at this the other week but got distracted.

Change looked good as far as I could tell and performance seemed unchanged, as expected, so should be all good to merge.

PR #439, opened 2022-10-26
Add max batch size limit to BatchEventProcessor

### LINKED ISSUE #438: Add a maximum batch size to the `BatchEventProcessor`
To limit the number of events processed by `BatchEventProcessor` before updating the sequence currently you can implement `EventHandler::setSequenceCallback` to store the `Sequence` and update it when needed while handling events. See `com.lmax.disruptor.examples.EarlyReleaseHandler` for an example.

However, this doesn't play well with the concept of Rewinding in the `BatchEventProcessor`, see #437 so instead we should add an optional constructor argument to the `BatchEventProcessor` for setting a maximum batch size.

Palmr:
Add an optional `maxBatchSize` parameter to `BatchEventProcessor`.

Defaults to `Integer.MAX_VALUE`.

Limits the amount of events that will be processed before updating the sequence in the `BatchEventProcessor`.

Closes #438

### DIFF
--- a/src/main/java/com/lmax/disruptor/BatchEventProcessor.java
+++ b/src/main/java/com/lmax/disruptor/BatchEventProcessor.java
@@ -33,6 +33,7 @@ public final class BatchEventProcessor<T>
     private static final int IDLE = 0;
     private static final int HALTED = IDLE + 1;
     private static final int RUNNING = HALTED + 1;
+    private static final int DEFAULT_MAX_BATCH_SIZE = Integer.MAX_VALUE;
 
     private final AtomicInteger running = new AtomicInteger(IDLE);
     private ExceptionHandler<? super T> exceptionHandler;
@@ -86,7 +87,7 @@ public final class BatchEventProcessor<T>
             final EventHandler<? super T> eventHandler
     )
     {
-        this(dataProvider, sequenceBarrier, eventHandler, Integer.MAX_VALUE);
+        this(dataProvider, sequenceBarrier, eventHandler, DEFAULT_MAX_BATCH_SIZE);
     }
 
     @Override
--- a/src/main/java/com/lmax/disruptor/BatchEventProcessor.java
+++ b/src/main/java/com/lmax/disruptor/BatchEventProcessor.java
@@ -196,9 +196,9 @@ public final class BatchEventProcessor<T>
                     final long availableSequence = sequenceBarrier.waitFor(nextSequence);
                     final long endOfBatchSequence = min(nextSequence + batchLimitOffset, availableSequence);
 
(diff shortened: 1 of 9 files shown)


src/main/java/com/lmax/disruptor/BatchEventProcessor.java
```diff
@@ -49,19 +51,44 @@
...
+     * @param sequenceBarrier on which it is waiting.
+     * @param eventHandler    is the delegate to which events are dispatched.
+     */
+    public BatchEventProcessor(
+            final DataProvider<T> dataProvider,
+            final SequenceBarrier sequenceBarrier,
+            final EventHandler<? super T> eventHandler
+    )
+    {
+        this(dataProvider, sequenceBarrier, eventHandler, Integer.MAX_VALUE);
```
swarren12, line 89: Can we replace this `Integer.MAX_VALUE` with a constant? `DEFAULT_MAX_BATCH_SIZE` maybe?

src/test/java/com/lmax/disruptor/MaxBatchSizeEventProcessorTest.java
```diff
@@ -0,0 +1,128 @@
...
+    public void shouldLimitTheBatchToConfiguredMaxBatchSize() throws Exception
+    {
+        long sequence = 0;
+        for (int i = 0; i < PUBLISH_COUNT; i++)
+        {
+            sequence = ringBuffer.next();
+        }
+        ringBuffer.publish(sequence);
+
+         //Wait for consumer to process all events
```
swarren12, line 67: Given that we do this more than once, and we have a duplicated comment, maybe it'd be worth replacing this with a method?

swarren12: Looking a bit more, I think these might be the only uses of `countDownLatch` outside `BatchLimitRecordingHandler` (if I'm wrong, I blame Github's diff UI 😁) and so maybe the whole latch could be encapsulated in that class?

sea36:
This changes the behaviour of the call on onBatchStart().  Previously it would receive notification of the entire count of pending events, now it receives the size of the batch that is going to be processed.  That does fit with the javadoc for the method, but the old value was useful for monitoring purposes - it is useful the know the size of the pending queue of events.

ocoanet:
In my view, `BatchEventProcessor` is almost an implementation detail for many disruptor users. The processor is often instantiated through the DSL API. Is there a plan to expose the max batch size in the DSL API? It could be added as a default member in `EventHandler<T>`.

swarren12 (approved):

Issue #397, opened 2021-10-13
Why does `TimeoutHandler` get to throw `Exception`?


Palmr:
`TimeoutHandler::onTimeout` throws a checked exception if the implementation is unable to handle the timeout.

`BatchEventProcessor::notifyTimeout` catches `Throwable`, so no need for checked exception type.

The exception gets passed to the exception handler which may handle or bail out of the `BatchEventProcessor::processEvents` depending on the `ExceptionHandler` implementation.

This is leaning towards use of exceptions for control flow and perhaps we could look at returning something useful from `TimeoutHandler::onTimeout` instead?

SteafanMrZhou:
I think the ExceptionHandler is only leaning towards to graph all type exceptions and distributing to exception handler by excetion type.
Liking this way,`TimeoutHandler::onTimeout` can deal with itself exception and return something useful information about timeout:waitTime,isInterrupted,isDeadLock,etc.
Anykinds exception will deal with like this in disruptor.

grumpyjames:
There's a further point raised from https://github.com/LMAX-Exchange/disruptor/issues/429 which makes me think "isn't the whole of timeout handling massively exception-as-control-flow ?

PR #440, opened 2022-10-26
Rewindable event handler separation

### LINKED ISSUE #437: `EventHandler::setSequenceCallback` will not play well with `RewindableException` throwing event handlers
Since #364 we have the concept of `EventHandler::onEvent` being able to throw a `RewindableException` which the `BatchEventProcessor` will handle and reset the sequence to the start of batch before trying to process the events again.

With the existing concept of `EventHandler::setSequenceCallback` an `EventHandler` can get a reference to the sequence and move it on of its own accord. Common examples for wanting to do this would be limiting the batch size (See `com.lmax.disruptor.examples.EarlyReleaseHandler`) or updating the sequence after flushing data to IO so downstream consumers can be unlocked.

If an `EventHandler` could throw `RewindableException` after it had updated the sequence by using `EventHandler::setSequenceCallback` then things would not go well.

**Expected behaviour**
We should make the option to use `EventHandler::setSequenceCallback` or throw `RewindableException` mutually exclusive. The `BatchEventProcessor` should not catch and handle `RewindableException` if the `EventHandler` implements `EventHandler::setSequenceCallback`.

Palmr:
This should give us type-safe rewindable and non-rewindable `EventHandler` implementations.

Non-rewindable implementations will not be able to throw `RewindableException` but will be able to implement `EventHandler::setSequenceCallback` and the opposite for rewindable implementations.

This will fix #437


src/main/java/com/lmax/disruptor/BatchEventProcessor.java
```diff
@@ -37,31 +38,112 @@
     private ExceptionHandler<? super T> exceptionHandler;
     private final DataProvider<T> dataProvider;
     private final SequenceBarrier sequenceBarrier;
-    private final EventHandler<? super T> eventHandler;
+    private final BaseEventHandler<? super T> eventHandler;
+    private final int batchLimitOffset;
     private final Sequence sequence = new Sequence(Sequencer.INITIAL_CURSOR_VALUE);
     private BatchRewindStrategy batchRewindStrategy = new SimpleBatchRewindStrategy();
     private int retriesAttempted = 0;
+    private final boolean rewindable;
```
swarren12, line 46: Rather than a `boolean`, could we have behaviour?

swarren12: Thinking something like adding an private inner interface:

```java
private final interface RewindBehaviour
{
      long onRewindableException(RewindableException e);
}
```

and then there could be two private methods:

```java
private long beKindRewind(final RewindableException e)
{
    if (this.batchRewindStrategy.handleRewindException(e, ++retriesAttempted) == REWIND)
    {
        return startOfBatchSequence;
    }
    else
    {
        retriesAttempted = 0;
        throw e;
    }
}

private long rewindNotAllowed(final RewindableException e)
{
    throw new RuntimeException("Rewindable Exception thrown from a non-rewindable event handler", e);
}
```

Constructors would then take `this::beKindRewind` in place of `true` and `this::rewindNotAllowed` in place of false? (one catch here is that they might not be referencable in `this()` calls, so eh?).

This would save continuously checking an `if` that will never change; on the other hand it does possibly make navigation slightly harder?

Also, this really only works because `retriesAttempted` appears to be instance scoped? I'm assuming this is so we persist failures across calls until we get a success, but it's not entirely clear to me?

src/main/java/com/lmax/disruptor/BatchEventProcessor.java
```diff
@@ -38,40 +38,78 @@
...
+        this.batchLimitOffset = maxBatchSize - 1;
+
+        this.rewindable = rewindable;
+    }
 
     /**
      * Construct a {@link EventProcessor} that will automatically track the progress by updating its sequence when
      * the {@link EventHandler#onEvent(Object, long, boolean)} method returns.
      *
+     * <p>This constructor will not support rewinding batches.
```
grumpyjames, line 73: How about 'The created BEP will not support batch rewind". The constructor is just a constructor

src/main/java/com/lmax/disruptor/BatchEventProcessor.java
```diff
@@ -38,40 +38,78 @@
...
+        this(dataProvider, sequenceBarrier, eventHandler, maxBatchSize, false);
 
         eventHandler.setSequenceCallback(sequence);
     }
 
+    /**
+     * Construct a {@link EventProcessor} that will automatically track the progress by updating its sequence when
+     * the {@link EventHandler#onEvent(Object, long, boolean)} method returns.
+     *
+     * <p>This constructor will support rewinding batches.
```
grumpyjames, line 96: Same same - the constructor isn't supporting anything!

src/main/java/com/lmax/disruptor/BaseEventHandler.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.lmax.disruptor;
+
+interface BaseEventHandler<T>
```
grumpyjames, line 18: Gentle preference for EventHandlerBase over BaseEventHandler, which sounds like an EventHandler that someone has done something unpleasant to.

grumpyjames (approved):
I sort of hate that we've got three constructors on BEP; if it weren't such a ballache I'd be suggesting some named factory methods instead.

I agree with Simon's suggestion of behaviour rather than boolean.

Don't think any of my comments are showstoppers.

sea36:
`RewindableException extends Throwable`

Extending `Throwable` :(
https://stackoverflow.com/a/2756001

And if we're going that path, it isn't an `Exception` any more.

swarren12:
> `RewindableException extends Throwable`
> 
> Extending `Throwable` :( https://stackoverflow.com/a/2756001
> 
> And if we're going that path, it isn't an `Exception` any more.

I also questioned this offline - my understanding of the reasoning here is that we have no way of allowing the regular handlers to throw an arbitrary exception _but not a `RewindableException`_. Making it an `Error` wouldn't work, as these are unchecked, so the only other option would be to change the signature of `onEvent()` in `EventHandler`.

Something like this would work in theory:

```
void onEvent(T event, long sequence, boolean endOfBatch) throws DisruptorException;
```

but that would require catching and wrapping every `Exception`, and would probably make migration onto 4.0 painful.

Palmr:
On the topic of exceptions and use of Throwable: I think it's justified here as it's a specific exception we want to raise in a certain circumstance.

There's probably some major refactoring I'd like to make in a future major release to make `EventHandler::onEvent` return something as right now we're straying down the path of exceptions for control flow. Another case in #397 

For now I think it's a fair trade off as the `RewindableException` is not something that should bubble out past the `BatchEventProcessor` and concerns of it not being caught by those only catching Exception *should* be mostly moot in this case.

swarren12 (approved):
LGTM

Wonder if we can move the `retriesAttempted` variable off the handler into the retry strategy as a next-step refactor?

grumpyjames:
Man this interface is _appalling_ for reviewing.

PR #445, opened 2022-12-24
jcstress 0.15

sullis:
https://mail.openjdk.org/pipermail/jcstress-dev/2021-December/000965.html

https://mvnrepository.com/artifact/org.openjdk.jcstress/jcstress-core

### DIFF
--- a/gradle/jcstress.gradle
+++ b/gradle/jcstress.gradle
@@ -3,7 +3,7 @@
 //
 
 jcstress {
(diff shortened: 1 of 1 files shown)


Palmr:
Not really suitable for a PR
(Also, there's a reason it was pinned at 0.11 still, later versions make CI checks exceedingly slow and this PR does nothing to address that)

Issue #456, opened 2023-09-15
Directly replace Fence with the methods setRelease, setVolatile, and getAcquire. No need to set barriers via VarHandle


chenggwang:
**Describe the bug**
This can't be considered a bug, just some minor use of VarHandle's API instead of fence. It's not a high priority either.

**Expected behavior**
Cleaner code and clearer semantics. Avoid using fences at the application level if possible, unless the existing read/write APIs are not sufficient

**Desktop (please complete the following information):**
 - OS: [windows]
 - Version [win 11]
 - JVM Version [openjdk 17]

**Additional context**
There are two files involved Sequence.java
SequenceVarHandleBarrier.java, read and write methods

chenggwang:
@Palmr By the way . I see that `handleEventsWithWorkerPool` has been removed. Are there any other APIs in the new version that can directly achieve non-duplicate consumption? If not, what should I do?

Palmr:
As mentioned on the PR that followed this: There is seemingly no tangible benefit, and the alternative proposed was already considered.

Closing this issue, though it could be re-opened if something were to change.

PR #457, opened 2023-09-15
#456 Directly replace Fence with the methods setRelease, setVolatile, and getAcquire. No need to set barriers via VarHandle

chenggwang:
Since we already have VALUE_FIELD, it is more semantically clear to directly replace Fence with the methods setRelease, setVolatile, and getAcquire. No need to set barriers via VarHandle.It's not a high priority, and it's not really a bug

### DIFF
--- a/src/main/java/com/lmax/disruptor/Sequence.java
+++ b/src/main/java/com/lmax/disruptor/Sequence.java
@@ -69,13 +69,12 @@ public class Sequence extends RhsPadding
 
     /**
      * Create a sequence with a specified initial value.
-     *
+     * Perform an ordered write of this sequence.
      * @param initialValue The initial value for this sequence.
      */
     public Sequence(final long initialValue)
     {
-        VarHandle.releaseFence();
-        this.value = initialValue;
+        VALUE_FIELD.setRelease(this,initialValue);
     }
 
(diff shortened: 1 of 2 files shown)


chenggwang:
@Palmr I don't think it's going to cause a barrier loss. Do you think that's okay?

chenggwang:
Can anyone help review this?Thanks！

chenggwang:
> There's already a version of this in the code for benchmarking, see [SequenceVarHandle.java](https://github.com/LMAX-Exchange/disruptor/blob/95c705f60c1833b07f1fed6e08a08d7bee7f0971/src/test/java/com/lmax/disruptor/alternatives/SequenceVarHandle.java)
> 
> As noted in [SequenceBenchmark.java](https://github.com/LMAX-Exchange/disruptor/blob/95c705f60c1833b07f1fed6e08a08d7bee7f0971/src/jmh/java/com/lmax/disruptor/SequenceBenchmark.java#L311) we felt the manual barriers were nicer since they cut down on boxing and casting.
> 
> As you say, "It's not a high priority, and it's not really a bug" so it's unlikely this PR will be merged unless some befit can be found and demonstrated.

OK

PR #461, opened 2023-10-04
Unify EventHandler implementations.

Hc747:
### Motivation
The changes released in `4.0.0` are breaking compared to those available in `4.0.0.RC1` and mean that EventHandler's can _either_ receive a sequence callback _or_ have rewindable logic, but not both. These changes amalgamate the APIs so that an EventHandler can be _both_ rewindable and able to receive sequence callbacks.

### Changes
- Remove EventHandlerBase component.
- Remove RewindableEventHandler component.
- Amalgamate and unify APIs of EventHandler, RewindableEventHandler and EventHandlerBase.
- Revert RewindableException to RuntimeException instead of Throwable.
- Denote BatchEventProcessor as being rewindable if internal RewindHandler is not an instance of NoRewindHandler.
- Update DSL and builder API to account for above changes.
- Update relevant test cases.


Palmr:
You are correct, "EventHandler's can either receive a sequence callback or have rewindable logic, but not both".
This is for good reason and I would question anyone wanting the ability to do both.

I'd suggest reading the PR that separated them https://github.com/LMAX-Exchange/disruptor/pull/440 and importantly the associated Issue https://github.com/LMAX-Exchange/disruptor/issues/437

Adding rewinding to the Batch Event Processor was a new feature for v4.
After the RC we realised it would be a _very bad idea_ to have the ability to move the sequence on when the Batch Event Processor may rewind.

The fix for this that we implemented was to separate the implementations.
This should mean rewindability and getting a hook to the sequence have a separation enforced at compile time rather than runtime exceptions or the need to read documentation to find out why doing both would be a bad idea.

Since 4.0.0-RC1 was an RC, we were fine with adding in a breaking change atop that considering that it removed such an obvious footgun.

Especially as it was only a breaking change for anyone who'd jumped on the new-in-v4 feature of rewindable Batch Event Processor, the impact is minimal, and in a release full of breaking changes from 3.x anyway.

Issue #449, opened 2023-02-23
Can a new 3.X version be released based on 3.4.4?


724027069:
Version 3.4.4 is released in 2021. Can you release a new version without any changes? If the version is not released, people will think that the version is inactive or vulnerabilities cannot be fixed in time. Thank you.

Palmr:
> Can you release a new version without any changes?

Technically, yes. Would we? No.

> If the version is not released, people will think that the version is inactive or vulnerabilities cannot be fixed in time.

I would hope people would only think this if there were any vulnerabilities. If there was a vulnerability in the 3.x version we would release a new version with a fix.

No releases will ever be done for the sake of bumping up a version number.

adski31:
The (test) dependency junit:junit:4.12 is vulnerable as per [Maven Central](https://mvnrepository.com/artifact/junit/junit/4.12). Spin a 3.4.5 with junit 4.13.2 and I bet you'd make a lot of people clamoring for a new revision very happy.

Palmr:
I'd hope anyone who might be using the Disruptor would be able to distinguish between a dependency and a test dependency.

For people impacted by that test dependency having a CVE, that's developers of the 3.x Disruptor branch, I would expect they could read the CVE too and understand it's about the TemporaryFolder rule, which isn't used in any Disruptor tests. So moot.

I think it would likely annoy a lot of people seeing a new release go out with zero functional changes.

PR #470, opened 2024-01-03
Fix CVE-2020-15250 in 3.4.X

NathanD001:
https://security.snyk.io/vuln/SNYK-JAVA-JUNIT-1017047

Upgrading to Disruptor 4.0.X is difficult because it is not compatible with Log4j2 2.22.X
https://github.com/apache/logging-log4j2/issues/1829

I think it would be good to have a 3.4.5 release that has this CVE addressed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,7 @@ repositories {
 
 dependencies {
     checkstyle 'com.puppycrawl.tools:checkstyle:8.2'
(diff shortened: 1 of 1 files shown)


Palmr:
https://github.com/LMAX-Exchange/disruptor/issues/449#issuecomment-1442395506

From the last time this was mentioned

Issue #154, opened 2016-05-25
if sequence beyond long range?


kkzzzzzz:
when we publish Event  the  sequence will increase , the sequence while  beyond long range?

mikeb01:
At 1,000,000,000 messages a second that will take around 270 years.  I don't think it is a issue.

PR #475, opened 2024-07-18
Fix:add a testcase to inform library user that here exists a risk of …

TenzT:
Fix:add a testcase to inform library user that here exists a risk of arithmetic overflow in Sequence once events exceeds Long.MAX_VALUE.


swarren12:
There appear to be no changes in this PR. Is this intentional?

isolgpus:
@mikeb01 response in #154 sums it up nicely. 
The disruptor does not support the long overflow case for sequence numbers as it would take around 270 years at 1 billion events per second.

PR #479, opened 2024-07-30
Refactored Identifiers

congyuluo:
Refactored identifiers to more appropriate names according to its usage in the code. This is to improve code clarity.

### DIFF
--- a/src/test/java/com/lmax/disruptor/alternatives/MultiProducerSequencerVarHandle.java
+++ b/src/test/java/com/lmax/disruptor/alternatives/MultiProducerSequencerVarHandle.java
@@ -112,9 +112,9 @@ public final class MultiProducerSequencerVarHandle extends AbstractSequencer
      * @see Sequencer#next(int)
      */
     @Override
-    public long next(final int n)
+    public long next(final int numberOfEntities)
     {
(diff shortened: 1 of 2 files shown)


src/test/java/com/lmax/disruptor/alternatives/MultiProducerSequencerVarHandle.java
```diff
@@ -112,9 +112,9 @@ public long next()
      * @see Sequencer#next(int)
      */
     @Override
-    public long next(final int n)
+    public long next(final int numberOfEntities)
```
ocoanet, line 115: I think this renaming is very suspicious. Why did you rename the parameter here and not in the interface and other implementing types? Also, why did you rename `n` in `next` and not in `tryNext`? Moreover, `numberOfEntities` is a totally generic name that is not related to the Disruptor: `numberOfSequences` or `numberOfEvents` would have been better. Anyway, I do not think renaming this parameter is a good idea.

ocoanet:
It looks like an automated PR or a change for the sake of change.

swarren12:
> It looks like an automated PR or a change for the sake of change.

Agree. Also it's a file under `src/test`. Going to close this.

PR #480, opened 2024-08-18
fix: Message loss

2498298782Zql:
Fix the issue where messages are lost if a consumer that is processing an event is killed.


Palmr:
I think for this PR to get anywhere you're going to want to add at least one test to demonstrate your change working.

PR #483, opened 2024-08-31
Add JMH benchmark to measure multi-producer batch publication performance.

nicholassm:
Batch publication has much more complicated logic in the `MultiProducerSequencer` compared to the `SingleProducerSequencer` where the cursor is simply updated with the high sequence value.
Therefore, there is also room for optimizations! :-)
Specifically, I am working on an idea to use just a single bit per slot in the ring buffer to represent "publication status" rather than an `int` and it should have the same performance for single event publication but better performance for the batch publication case.
I've already seen great results with this approach in the Rust port I've written (https://github.com/nicholassm/disruptor-rs) and I would like to explore the same idea in the "original" Disruptor.
But first it would be nice with a JMH benchmark to actually measure things.
That's what this small Pull-Request adds. On my laptop, batch publication has 3 times higher throughput compared to single event publication. Can we do better?


nicholassm:
Any progress on merging this one?

Kind regards
Nicholas

grumpyjames:
This looks very exciting - apologies it took us so long to wake up.

Merged - looking forward to seeing the number go up :-)

PR #500, opened 2025-05-18
Producer wait strategy

ryntric:
This pull request adds an ability to specify a producer wait policy.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ plugins {
 }
 
 group = 'com.lmax'
-version = new Version(major: 4, minor: 0, revision: 0, snapshot: true)
+version = new Version(major: 5, minor: 0, revision: 0, snapshot: false)
 
 defaultTasks 'build'
 
--- a/src/docs/asciidoc/en/changelog.adoc
+++ b/src/docs/asciidoc/en/changelog.adoc
@@ -4,6 +4,9 @@
 
 = Changelog
 
+== 5.0.0
+* `ProducerWaitStrategy` is now supported.
+
 == 4.0.0
 
 * Minimum Java version now 11
--- a/src/examples/java/com/lmax/disruptor/examples/MultiProducerWithTranslator.java
+++ b/src/examples/java/com/lmax/disruptor/examples/MultiProducerWithTranslator.java
@@ -1,5 +1,6 @@
 package com.lmax.disruptor.examples;
 
+import com.lmax.disruptor.BlockingProducerWaitStrategy;
 import com.lmax.disruptor.BlockingWaitStrategy;
 import com.lmax.disruptor.EventFactory;
 import com.lmax.disruptor.EventHandler;
@@ -69,7 +70,7 @@ public class MultiProducerWithTranslator
     {
         Disruptor<ObjectBox> disruptor = new Disruptor<>(
                 ObjectBox.FACTORY, RING_SIZE, DaemonThreadFactory.INSTANCE, ProducerType.MULTI,
-                new BlockingWaitStrategy());
+                new BlockingWaitStrategy(), new BlockingProducerWaitStrategy());
         disruptor.handleEventsWith(new Consumer()).then(new Consumer());
         final RingBuffer<ObjectBox> ringBuffer = disruptor.getRingBuffer();
         Publisher p = new Publisher();
--- a/src/jcstress/java/com/lmax/disruptor/LoggerInitializationStress.java
+++ b/src/jcstress/java/com/lmax/disruptor/LoggerInitializationStress.java
@@ -74,7 +74,8 @@ public class LoggerInitializationStress
                 128,
                 Executors.defaultThreadFactory(),
                 ProducerType.MULTI,
-                new BlockingWaitStrategy());
+                new BlockingWaitStrategy(),
+                new BlockingProducerWaitStrategy());
         disruptor.setDefaultExceptionHandler(SimpleEventExceptionHandler.INSTANCE);
         disruptor.handleEventsWith(SimpleEventHandler.INSTANCE);
         disruptor.start();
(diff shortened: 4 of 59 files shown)


ryntric:
What shall I do to fix this step?

adacooke:
Hi,
Thanks for opening a PR. It looks like your change mostly came from the TODOs in SingleProducerSequencer and MultiProducerSequencer. That todo was added 11 years ago, and Disruptor has been through many maintainers since, could I ask what the use case is for this, why do you want to change the wait strategy when the ring buffer is full?

ryntric:
> Hi, Thanks for opening a PR. It looks like your change mostly came from the TODOs in SingleProducerSequencer and MultiProducerSequencer. That todo was added 11 years ago, and Disruptor has been through many maintainers since, could I ask what the use case is for this, why do you want to change the wait strategy when the ring buffer is full?

Hi, sure. I am developing my own library that is built upon Disruptor and want to have an ability to specify a timeout for adding a task. For example, we read messages from Kafka topics, and we want to handle them concurrently via Disruptor for low latency. Those messages are stateful, and we can drop/skip this message because the next one will contain a full state.

adacooke:
> Hi, sure. I am developing my own library that is built upon Disruptor and want to have an ability to specify a timeout for adding a task. For example, we read messages from Kafka topics, and we want to handle them concurrently via Disruptor for low latency. Those messages are stateful, and we can drop/skip this message because the next one will contain a full state.

So if I'm understanding you correctly, you're consuming from Kafka, and putting the messages on a Disruptor. Then when that ring buffer is full, you want to spin for a while, and then drop messages? This is a fairly significant API change, would overflow strategies like we use in [Disruptor Proxy](https://github.com/LMAX-Exchange/disruptor-proxy/blob/master/src/main/java/com/lmax/tool/disruptor/OverflowStrategy.java) be a viable alternative to implement outside of Disruptor?

Also, because you're changing code on one of the hottest paths in Disruptor, could you please post the results of the benchmarks before and after your change?

ryntric:
> > Hi, sure. I am developing my own library that is built upon Disruptor and want to have an ability to specify a timeout for adding a task. For example, we read messages from Kafka topics, and we want to handle them concurrently via Disruptor for low latency. Those messages are stateful, and we can drop/skip this message because the next one will contain a full state.
> 
> So if I'm understanding you correctly, you're consuming from Kafka, and putting the messages on a Disruptor. Then when that ring buffer is full, you want to spin for a while, and then drop messages? This is a fairly significant API change, would overflow strategies like we use in [Disruptor Proxy](https://github.com/LMAX-Exchange/disruptor-proxy/blob/master/src/main/java/com/lmax/tool/disruptor/OverflowStrategy.java) be a viable alternative to implement outside of Disruptor?
> 
> Also, because you're changing code on one of the hottest paths in Disruptor, could you please post the results of the benchmarks before and after your change?

Unfortunately, `Disruptor Proxy` does not suit me. Benchmarks have been attached below. They are separated by branch.

| 									producer_wait_strategy		  	|			Mode | Cnt | Score Error     | Units |
|-------------------------------------------------------------------|----------------|-----|-----------------|-------|
| `SingleProducerWithBlockingProducerWaitStrategy.producing`  	  	|			avgt | 	5  | 5.807 ± 0.273	 | ns/op |
| `SingleProducerWithTimeoutBlockingProducerWaitStrategy.producing` | 			avgt |   5  | 31.131 ± 1.402 | ns/op |
| `MultiProducerWithBlockingProducerWaitStrategy.producing`  		|			avgt |  5  | 65.253 ± 6.222  | ns/op |
| `MultiProducerWithTimeoutBlockingProducerWaitStrategy.producing`  |			avgt |   5  |108.157 ± 2.899 | ns/op |



| 									master						  	|			Mode | Cnt | Score Error     | Units |
|-------------------------------------------------------------------|----------------|-----|-----------------|-------|
| `SingleProducerBenchmark.producing`  	  							|			avgt | 	5  | 6.354 ± 1.104   | ns/op |
| `MultiProducerBenchmark.producing`								|			avgt |	5  | 81.374 ± 34.055 |  ns/op|

adacooke:
Thanks for posting these benchmarks, and good spot on the nanoTime call. We've discussed this PR internally and due to the niche use case and breaking nature, we're not going to merge this, although we'd encourage you to maintain this as a fork. 

If you do want to maintain this, I've ran some of the end to end perf tests locally, and there's a pretty large drop in performance in the throughput benchmarks (I measure 4.2 billion ops/second on master, and 2.8 billion ops/second on your branch), I haven't investigated this in too much detail, but this might be related to the code in the sequencers being moved around.

ryntric:
> Thanks for posting these benchmarks, and good spot on the nanoTime call. We've discussed this PR internally and due to the niche use case and breaking nature, we're not going to merge this, although we'd encourage you to maintain this as a fork.
> 
> If you do want to maintain this, I've ran some of the end to end perf tests locally, and there's a pretty large drop in performance in the throughput benchmarks (I measure 4.2 billion ops/second on master, and 2.8 billion ops/second on your branch), I haven't investigated this in too much detail, but this might be related to the code in the sequencers being moved around.

Thanks for the answer. Have you noticed performance decreasing on a new logic that is related to either TimeoutBlockingProducerWaitStrategy or BlockingProducerWaitStrategy?

adacooke:
> Thanks for the answer. Have you noticed performance decreasing on a new logic that is related to either TimeoutBlockingProducerWaitStrategy or BlockingProducerWaitStrategy?

The perf tests I ran were using BlockingProducerWaitStrategy, using the timeout one, I get around 280 million ops/second, so it's probably worth trying to avoid doing that nanotime call, maybe don't call it until you've been unable to get a slot in the ringbuffer?

ryntric:
> > Thanks for the answer. Have you noticed performance decreasing on a new logic that is related to either TimeoutBlockingProducerWaitStrategy or BlockingProducerWaitStrategy?
> 
> The perf tests I ran were using BlockingProducerWaitStrategy, using the timeout one, I get around 280 million ops/second, so it's probably worth trying to avoid doing that nanotime call, maybe don't call it until you've been unable to get a slot in the ringbuffer?

The BlockingProducerWaitStrategy has a dummy implementation for getClaimed() and returns 0. The performance drop was rather caused by `hasAvailableCapacity(final int requiredCapacity, final boolean doStore)` in a loop. Every time this method is called, he stores the cursor value `cursor.setVolatile(nextValue);  // StoreLoad fence`.

PR #508, opened 2026-02-05
Add producer wait strategies for capacity backoff

svrohith9:
## Summary
  - add producer-side wait strategy API and implementations
  - wire producer backoff into single/multi sequencers
  - add RingBuffer/Disruptor overloads and docs updates
  - add unit tests for producer wait strategies

  ## Testing
  - ./gradlew test

### DIFF
--- a/src/docs/asciidoc/en/user-guide/10_using_the_disruptor.adoc
+++ b/src/docs/asciidoc/en/user-guide/10_using_the_disruptor.adoc
@@ -314,7 +314,8 @@ public class LongEventMain
             bufferSize,
             DaemonThreadFactory.INSTANCE,
             ProducerType.SINGLE, // <1>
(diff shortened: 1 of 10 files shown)


adacooke:
Hi,
Thank you for your PR! Can I ask what the intended use case is for this feature? Generally the ring buffer being full is a really bad case and I don't know what use case would require configuring how that is handled.

svrohith9:
@adacooke Thanks! The goal isn’t to normalize the full buffer. Those are still a bad sign. But brief
  ring‑full periods do happen (bursts, GC pauses, downstream hiccups). Today producers always
  parkNanos(1), which hard‑codes one latency/CPU tradeoff. This change just lets apps choose a
  different backoff policy for those short contention windows. Default behavior stays the same.

PR #514, opened 2026-08-15
fix(dsl): remove dead shutdown exception handler in Disruptor.shutdown()

### LINKED ISSUE #501: `handleOnShutdownException` might never be invoked in `shutdown`
I identified a potential flaw in the implementation of `Disruptor.shutdown`:
```java
public void shutdown()
{
    try
    {
        shutdown(-1, TimeUnit.MILLISECONDS);
    }
    catch (final TimeoutException e)
    {
        exceptionHandler.handleOnShutdownException(e);
    }
}
```

The `-1` parameter is interpreted as an infinite timeout, which aligns with the method's comments. As a result, `TimeoutException` will never be thrown and the `handleOnShutdownException` callback will never be invoked.

To resolve this, I suggest removing the try / catch block.

If needed, I can create a pull request with the changes.

arimu1:
## Summary

- Removes unreachable `handleOnShutdownException` path from `Disruptor.shutdown()` — the no-arg overload delegated to `shutdown(-1, …)`, which never throws `TimeoutException`, so the catch block was dead code.
- Implements infinite-timeout backlog drain inline (same busy-spin behaviour as the timed overload with `-1`).
- Adds `shouldCompleteShutdownWithoutInvokingShutdownExceptionHandler` regression test in `DisruptorTest`.

Fixes #501

## Test plan

- [x] `./gradlew test --tests com.lmax.disruptor.dsl.DisruptorTest` (Java 21)
- [x] `./gradlew test --tests com.lmax.disruptor.ShutdownOnFatalExceptionTest` (Java 21)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/main/java/com/lmax/disruptor/dsl/Disruptor.java
+++ b/src/main/java/com/lmax/disruptor/dsl/Disruptor.java
@@ -381,14 +381,11 @@ public class Disruptor<T>
      */
     public void shutdown()
     {
-        try
+        while (hasBacklog())
         {
-            shutdown(-1, TimeUnit.MILLISECONDS);
+            // Busy spin
         }
-        catch (final TimeoutException e)
-        {
-            exceptionHandler.handleOnShutdownException(e);
-        }
+        halt();
     }
 
     /**
--- a/src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
+++ b/src/test/java/com/lmax/disruptor/dsl/DisruptorTest.java
@@ -616,23 +616,27 @@ public class DisruptorTest
     }
 
     @Test
-    public void shouldCompleteShutdownWithoutInvokingShutdownExceptionHandler()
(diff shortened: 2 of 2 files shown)


arimu1:
Addressed the second review:

1. **Stripped `Co-authored-by: Cursor` from all commits** — branch history rewritten via `commit-tree`; author/committer is `arimu1` only.

2. **Renamed and reframed the test** — `shouldShutdownDrainBacklogWithoutTimeoutException` now exercises the no-arg shutdown busy-spin with `DelayedEventHandler` + chained latch, without the vacuous `assertNull(exceptionHandled)` assertion. The test documents that `handleOnShutdownException` was unreachable dead code on master because `shutdown(-1, …)` never throws `TimeoutException`; both old and new implementations drain backlog the same way, so there is no meaningful before/after failure mode for the removed catch block.

3. **Commit messages** clarify this is dead-code removal with drain coverage, not a behaviour change.

